PR #39, opened 2018-12-11
Integrating with travis-ci

neoword:
* Fixes Issue #6
* Integrates with travis-ci

## Added
* Integration with travis-ci 

# PR Checklist Forms

- [x] CHANGELOG.md updated
- [x] Reviewer assigned
- [x] PR assigned (presumably to submitter)
- [x] Labels added (enhancement, bug, documentation) 
- [x] Milestone selected
- [x] Any annotations for reviewers have been added
      (with preference of annotating directly in code comments)
- [x] PR builds successfully

PR #41, opened 2018-12-11
Schema Validation support via `SchemaManager` interface 

westeras:
# stream-registry PR

Adding an interface for defining schema validation/registration

## Added
* `SchemaManager` interface
* Added ability to use custom `SchemaManager` implementation, which is used to validate and register schema on stream upsert
* Default schema manager implementation `ConfluentSchemaManager` which is used against a Confluent Schema Registry instance
* Stream Registry /compatibility endpoint for validating schema compatibility when schemas change

# PR Checklist Forms

_(to be filled out by PR submitter)_
- [x] This template filled (above this section)
- [x] CHANGELOG.md updated
- [x] Reviewer assigned
- [x] Milestone selected
- [x] PR builds successfully


core/src/main/java/com/homeaway/streamingplatform/extensions/validation/SchemaRegistrar.java
```diff
@@ -0,0 +1,57 @@
...
+ */
+public interface SchemaRegistrar {
+
+    /**
+     * Used to validate the stream's key and value schemas
+     *
+     * @param stream object
+     * @return whether schemas are valid or not
+     */
+    boolean isSchemaValid(Stream stream);
```
prabhakar1983: Can we just pass the schema as string in the parameter.

also, it is preferred to use the same/similar signature or method names of the confluent's SchemaRegistry client
https://github.com/confluentinc/schema-registry/blob/master/client/src/main/java/io/confluent/kafka/schemaregistry/client/SchemaRegistryClient.java

core/src/main/java/com/homeaway/streamingplatform/extensions/validation/EmptySchemaRegistrar.java
```diff
@@ -0,0 +1,47 @@
...
+import com.homeaway.streamingplatform.exceptions.SchemaRegistrationException;
+import com.homeaway.streamingplatform.model.Stream;
+
+@SuppressWarnings("unused")
+@Slf4j
+public class EmptySchemaRegistrar implements SchemaRegistrar {
+
+    @Override
+    public boolean isSchemaValid(Stream stream) {
+        return false;
```
prabhakar1983: I believe we will have a sub-module `confluent-schema-registry-validator` in this project, which has the concrete implementation for communicating with SchemaRegistry.

However, most other companies may not have SchemaRegistry setup, and might be using the EmptySchemaRegistrar implementation to make the app run.

In that case, do you think, `isSchemaValid` should always return true so that the app will function.

core/src/test/resources/config-dev.yaml
```diff
@@ -5,6 +5,10 @@ streamValidatorConfig:
   properties:
     validator.invalid-id: 13
 
+schemaRegistrarConfig:
+  className: com.homeaway.streamingplatform.extensions.validator.SchemaRegistrarIT$ValidSchemaRegistrar
```
prabhakar1983: Once you have the real implemention of submodule `schema-registry-validator`, can you please replace the above class with the real class.

core/src/main/java/com/homeaway/streamingplatform/extensions/validation/SchemaRegistrar.java
westeras: My thought is that we are not assuming the use of Confluent's schema registry which is why we are creating the abstraction in the first place.  Two simple methods make more sense to me.  @vinayakponangi what do you think?

core/src/main/java/com/homeaway/streamingplatform/StreamRegistryApplication.java
```diff
@@ -48,10 +48,7 @@
 import org.rocksdb.BlockBasedTableConfig;
 import org.rocksdb.Options;
 
-import com.homeaway.streamingplatform.configuration.InfraManagerConfig;
-import com.homeaway.streamingplatform.configuration.StreamRegistryConfiguration;
-import com.homeaway.streamingplatform.configuration.StreamValidatorConfig;
-import com.homeaway.streamingplatform.configuration.TopicsConfig;
+import com.homeaway.streamingplatform.configuration.*;
```
neoword: we need to put coding standards into this project.

* **no wildcard imports** except for unit-test helper/matchers please.

core/src/main/java/com/homeaway/streamingplatform/StreamRegistryApplication.java
```diff
@@ -168,7 +166,10 @@ public void run(final StreamRegistryConfiguration configuration, final Environme
         StreamValidator streamValidator = loadValidator(configuration, httpClient, regionDao);
         Preconditions.checkState(streamValidator != null, "streamValidator cannot be null.");
 
-        StreamDao streamDao = new StreamDaoImpl(managedProducer, managedKStreams, env, regionDao, infraManager, kafkaManager, streamValidator);
+        SchemaRegistrar schemaRegistrar = loadSchemaRegistrar(configuration);
+        Preconditions.checkState(schemaRegistrar != null, "schemaRegistrar cannot be null.");
```
neoword: why not perform this check/initialization during initialization ?

core/src/main/java/com/homeaway/streamingplatform/StreamRegistryApplication.java
```diff
@@ -223,6 +224,32 @@ public static StreamValidator loadValidator(StreamRegistryConfiguration configur
         return null;
     }
 
+    public static SchemaRegistrar loadSchemaRegistrar(StreamRegistryConfiguration configuration) {
+        SchemaRegistrarConfig schemaRegistrarConfig = configuration.getSchemaRegistrarConfig();
+
+        if (schemaRegistrarConfig != null) {
+            String schemaRegistrarClass = schemaRegistrarConfig.getClassName();
+
```
neoword: Consider adding a precondition check here requiring the class.

core/src/main/java/com/homeaway/streamingplatform/StreamRegistryApplication.java
```diff
@@ -223,6 +224,32 @@ public static StreamValidator loadValidator(StreamRegistryConfiguration configur
         return null;
     }
 
+    public static SchemaRegistrar loadSchemaRegistrar(StreamRegistryConfiguration configuration) {
+        SchemaRegistrarConfig schemaRegistrarConfig = configuration.getSchemaRegistrarConfig();
+
+        if (schemaRegistrarConfig != null) {
```
neoword: Consider adding a precondition check here requiring the config.

core/src/main/java/com/homeaway/streamingplatform/StreamRegistryApplication.java
neoword: Also... if you add this as a post-condition check in `loadSchemaRegistrar` you don't need to do the check here.

* See: https://github.com/homeaway/stream-registry/pull/41/files#diff-a74858465e2f75329645cef3bd8c5ca3R230
* See: https://github.com/homeaway/stream-registry/pull/41/files#diff-a74858465e2f75329645cef3bd8c5ca3R232

core/src/main/java/com/homeaway/streamingplatform/configuration/SchemaRegistrarConfig.java
```diff
@@ -0,0 +1,34 @@
...
+import lombok.NonNull;
+
+@Data
+public class SchemaRegistrarConfig {
+
+    @Valid
+    @NonNull
+    String className;
+
+    Map<String, ?> properties;
```
neoword: why `?` type parameter?
Won't this be a `String` or `Object`?
If you make it an `Object` doesn't that simplify the need to copy the map in the `loadSchemaRegistrar` above?

core/src/main/java/com/homeaway/streamingplatform/db/dao/impl/StreamDaoImpl.java
```diff
@@ -15,12 +15,7 @@
  */
 package com.homeaway.streamingplatform.db.dao.impl;
 
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
-import java.util.Properties;
+import java.util.*;
```
neoword: * **no wildcard imports** except for unit-test helper/matchers please.

core/src/main/java/com/homeaway/streamingplatform/extensions/validation/SchemaRegistrar.java
neoword: Hmm... @westeras can we use this as an interface?
https://homeaway.github.io/stream-registry/architecture/#schema-manager-interface

If not, help me understand the current differences and why.

neoword:
Also this is blocked on #39 to verify the build is passing.

image

core/src/main/java/com/homeaway/streamingplatform/extensions/validation/SchemaRegistrar.java
westeras: @neoword yes...I didn't know that existed...

prabhakar1983: @neoword 
myself and Adam had discussion on this. 
so, the Interface `SchemaManager` at this point would have only 2 methods
```
public SchemaReference registerSchema(String subject, String schema)
                                        throws SchemaManagerException;
```
```
public boolean checkCompatibility(String subject, int version,
                                    String schema)
                                    throws SchemaException;```
Once we have use-cases for other methods, we would include the rest of the methods, and implement them.

Comment (westeras) [reply core/src/main/java/com/homeaway/streamingplatform/extensions/validation/SchemaRegistrar.java]: @neoword @prabhakar1983 another thing I just thought about: if we follow this contract and only provide subject/version/schema to these methods, we still need some mechanism for identifying the correct schema registry URL.  My original design passed the entire stream object into the schema manager interface, which had an infra manager instance, and then used the `vpcList` and `hint` from the stream to get the property URL from the infra manager.  I can think of two options to get around this: assume only one master schema registry URL, or provide a method via schema manager interface to lookup proper schema registry URL

File: core/src/main/java/com/homeaway/streamingplatform/configuration/SchemaManagerConfig.java [hunk 10]
```diff
@@ -0,0 +1,36 @@
...
+
+import javax.validation.Valid;
+
+import lombok.Data;
+import lombok.NonNull;
+
+@Data
+public class SchemaManagerConfig {
+
+    public static final String SCHEMA_REGISTRY_URL = "schema.registry.url";
```
cricket007: Use the existing constant - https://github.com/confluentinc/schema-registry/blob/4.1.2-post/avro-serializer/src/main/java/io/confluent/kafka/serializers/AbstractKafkaAvroSerDeConfig.java#L37

core/src/main/java/com/homeaway/streamingplatform/extensions/schema/CompatibilityLevel.java
```diff
@@ -0,0 +1,20 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.homeaway.streamingplatform.extensions.schema;
+
+public enum CompatibilityLevel {
+    BACKWARD_TRANSITIVE, FORWARD_TRANSITIVE, FULL_TRANSITIVE
```
cricket007: Can this not reuse this? https://github.com/confluentinc/schema-registry/blob/4.1.2-post/client/src/main/java/io/confluent/kafka/schemaregistry/avro/AvroCompatibilityLevel.java#L19

westeras: I was confused as to why the SchemaManager reference interface [here](https://homeaway.github.io/stream-registry/architecture/#schema-manager-interface) had fewer compatibility levels.  Until @neoword sees this I will switch it to use the existing one

core/src/main/java/com/homeaway/streamingplatform/db/dao/impl/StreamDaoImpl.java
cricket007: Is there a Maven Plugin somewhere that disallows this?

core/src/main/java/com/homeaway/streamingplatform/extensions/validation/SchemaRegistrar.java
neoword: My only point here, is lets not re-invent the wheel.  

The interface proposed above was shamelessly pulled from here:
https://docs.confluent.io/current/schema-registry/docs/api.html

I believe most (all?) use cases are covered by that.
If you can think of a specific use case not covered by that, lets talk offline.

confluent-schema-manager/src/main/java/com/homeaway/streamingplatform/extensions/schema/ConfluentSchemaManager.java
```diff
@@ -0,0 +1,62 @@
...
+            long id = schemaRegistryClient.register(subject, avroSchema);
+            long version = schemaRegistryClient.getLatestSchemaMetadata(subject).getVersion();
+            schemaReference = new SchemaReference(subject, id, version);
+        } catch (IOException e) {
+            log.error("caught an IOException while registering a new schema for subject '{}'", subject);
+            throw new SchemaManagerException(e);
+        } catch (RestClientException e) {
+            log.error("caught a RestClientException while registering a new schema for subject '{}'", subject);
+            throw new SchemaManagerException(e);
+        }
```
neoword: why differentiate ?

why not just :
```
} catch (RuntimeException e) {
...
}
```
is RestClientException checked exception?
If yes, then consider the folloing
```
} catch (RestClientException | RuntimeException exception) {...}
```

confluent-schema-manager/src/main/java/com/homeaway/streamingplatform/extensions/schema/ConfluentSchemaManager.java
```diff
@@ -0,0 +1,62 @@
...
+        org.apache.avro.Schema avroSchema = new org.apache.avro.Schema.Parser().parse(schema);
+        try {
+            return schemaRegistryClient.testCompatibility(subject, avroSchema);
+        } catch (IOException e) {
+            log.error("caught an IOException while checking compatibility for subject '{}'", subject);
+            throw new SchemaException(e);
+        } catch (RestClientException e) {
+            log.error("caught a RestClientException while checking compatibility for subject '{}'", subject);
+            throw new SchemaException(e);
+        }
```
neoword: same here (as above)

core/src/main/java/com/homeaway/streamingplatform/StreamRegistryApplication.java
```diff
@@ -137,7 +140,7 @@ public void run(final StreamRegistryConfiguration configuration, final Environme
                 throw new IllegalStateException("Could not load/configure Infra Manager class", e);
             }
         }
-        ManagedInfraManager managedInfraManager= new ManagedInfraManager(infraManager);
+        ManagedInfraManager managedInfraManager = new ManagedInfraManager(infraManager);
 
         environment.lifecycle().manage(managedProducer);
         environment.lifecycle().manage(managedKStreams);
```
neoword: add a comment as to why we explicitly do this here

core/src/main/java/com/homeaway/streamingplatform/db/dao/impl/StreamDaoImpl.java
```diff
@@ -120,12 +126,18 @@ public void upsertStream(Stream stream) {
             log.error("Stream '{}' is not valid", stream.getName());
         }
 
+        // register schemas
```
neoword: Add a TODO to support multi-type per stream (for event sourcing and clickstream support)

core/src/main/java/com/homeaway/streamingplatform/db/dao/impl/StreamDaoImpl.java
```diff
@@ -153,12 +165,16 @@ public void upsertStream(Stream stream) {
                 avroStream.setCreated(System.currentTimeMillis());
                 key = AvroStreamKey.newBuilder().setStreamName(avroStream.getName()).build();
             }
+
             verifyAndUpsertTopics(stream);
             kafkaProducer.log(key, avroStream);
             log.info("Stream upserted for {}", stream.getName());
         } catch (IllegalArgumentException e) {
             log.error("caught an illegal argument exception");
             throw new IllegalArgumentException(e);
```
neoword: why catch and rethrow illegalargumentexception?

core/src/main/java/com/homeaway/streamingplatform/db/dao/impl/StreamDaoImpl.java
```diff
@@ -153,12 +165,16 @@ public void upsertStream(Stream stream) {
...
+        } catch (SchemaManagerException e) {
+            log.error("caught an exception during schema registration", e);
```
neoword: why catch and rethrow just to log?
Don't we log at the top of the callchain? If not, we should.

core/src/main/java/com/homeaway/streamingplatform/extensions/schema/SchemaReference.java
```diff
@@ -0,0 +1,29 @@
...
+package com.homeaway.streamingplatform.extensions.schema;
+
+import lombok.AllArgsConstructor;
+import lombok.Data;
+
+@Data
+@AllArgsConstructor
+public class SchemaReference {
+
+    private String subject;
```
neoword: Love this!

Makefile
```diff
@@ -13,7 +13,7 @@ run:
 	java -Xmx2G -server -XX:+UseG1GC -XX:MaxGCPauseMillis=20 -XX:InitiatingHeapOccupancyPercent=35 -XX:+ExplicitGCInvokesConcurrent -Djava.awt.headless=true -jar assembly/target/stream-registry*SNAPSHOT.jar server config-dev.yaml
 
 debug:
-	java -Xmx2G -server -XX:+UseG1GC -XX:MaxGCPauseMillis=20 -XX:InitiatingHeapOccupancyPercent=35 -XX:+ExplicitGCInvokesConcurrent -Djava.awt.headless=true -agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=5105 -jar assembly/target/stream-registry*SNAPSHOT.jar server config-dev.yaml
+	java -Xmx2G -server -XX:+UseG1GC -XX:MaxGCPauseMillis=20 -XX:InitiatingHeapOccupancyPercent=35 -XX:+ExplicitGCInvokesConcurrent -Djava.awt.headless=true -agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=5105 -jar assembly/target/stream-registry*SNAPSHOT.jar server config-dev.yaml
```
neoword: please revert (looks like you pushed something you were debugging)

cricket007: Suggest a change to `suspend=${SCHEMA_REGISTRY_DEBUG_SUSPEND:-'n'}`

neoword: yeah. something like that ^^^ would be great!

neoword:
Also... please add unit tests & integration tests.
And rebase.

core/src/main/java/com/homeaway/streamingplatform/configuration/SchemaManagerConfig.java
westeras: done

core/src/main/java/com/homeaway/streamingplatform/db/dao/impl/StreamDaoImpl.java
westeras: not totally sure, removed it

Makefile
westeras: modified so that suspend is 'n' by default but running `STREAM_REGISTRY_DEBUG_SUSPEND=y make debug` will suspend

core/src/main/java/com/homeaway/streamingplatform/db/dao/impl/StreamDaoImpl.java
westeras: yes we do, I removed the superfluous catch block

core/src/main/java/com/homeaway/streamingplatform/db/dao/impl/StreamDaoImpl.java
westeras: done

confluent-schema-manager/src/main/java/com/homeaway/streamingplatform/extensions/schema/ConfluentSchemaManager.java
westeras: It's two separate checked exceptions, neither of which inherit from the other so I changed it to `(IOException | RestClientException e)`

confluent-schema-manager/src/main/java/com/homeaway/streamingplatform/extensions/schema/ConfluentSchemaManager.java
westeras: addressed

confluent-schema-manager/src/main/java/com/homeaway/streamingplatform/extensions/schema/ConfluentSchemaManager.java
westeras: ended up changing this to `catch (IOException | RestClientException | RuntimeException e) {` to include catching an unchecked exception thrown via registration failure

core/src/main/java/com/homeaway/streamingplatform/extensions/schema/CompatibilityLevel.java
neoword: @westeras, yes that one is good.

However, we want ours to be agnostic of others... so that other people can swap in schema registries.  By defining compatibility here, we can plug into the compatibility levels that underlying schema registries provide.

core/src/main/java/com/homeaway/streamingplatform/StreamRegistryApplication.java
neoword: any ideas on why we manage this here ? (vs setup configuration only) ?

neoword:
@prabhakar1983 
> There could be teams who does not use AVRO as message format (Example: JSON), and does not need registration/validation.
> Do we need to add a field `skipSchemaValidation` in `@PUT upsertStream` to allow skipping the validation if the user decides so.

For now we support AVRO only -- primarily because we are explicitly opinionated in "schema-first"/"schema-on-write" designs.

Open an issue to have a holistic solution to support JSON. (We can then update the design docs, and the stories should then be obvious).

Lets not block this effort for now.  If there is a concrete use case that is blocking integration, lets consider that directly.

neoword:
This will be blocked until #57 is fixed . (and #57 is blocked #58)

Issue #66, opened 2018-12-19
Stream Upsert Ignores `streamName` path param


westeras:
# Desired Behavior

https://github.com/homeaway/stream-registry/blob/080e487c9a763b4d1847fd72ebfe375793ef46e9/core/src/main/java/com/homeaway/streamingplatform/resource/StreamResource.java#L89-L92

The stream resource should validate the provided `streamName` path param against the stream name in the Stream entity.  In addition, currently any feign clients generated against the SR's swagger endpoint behave unpredictably due to literal `"{streamName}"` being left in the URL

# Benefits

* Adds a layer of protection/validation on stream upsert
* Fixes feign client behavior derived from swagger

neoword:
#69 Fixes this.

PR #69, opened 2018-12-20
Fixing unused stream name path param

westeras:
Addresses #66 

# stream-registry PR

Path param in PUT /v0/streams/{streamName} resource isn't annotated and used as a path param, causing problems downstream for clients whose APIs are generated via the stream registry swagger (in certain places curly braces are left unencoded in HTTP requests)

### Changed
* Added `streamName` as an annotated path param as second argument of stream upsert method in StreamResource

# PR Checklist Forms

- [x] CHANGELOG.md updated
- [x] Reviewer assigned

### DIFF
--- a/core/src/main/java/com/homeaway/streamingplatform/resource/StreamResource.java
+++ b/core/src/main/java/com/homeaway/streamingplatform/resource/StreamResource.java
@@ -19,11 +19,9 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
 
-import javax.ws.rs.BadRequestException;
 import javax.ws.rs.Consumes;
 import javax.ws.rs.DELETE;
 import javax.ws.rs.GET;
-import javax.ws.rs.InternalServerErrorException;
 import javax.ws.rs.PUT;
 import javax.ws.rs.Path;
 import javax.ws.rs.PathParam;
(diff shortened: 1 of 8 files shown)


core/src/main/java/com/homeaway/streamingplatform/resource/StreamResource.java
```diff
@@ -89,19 +85,29 @@ public StreamResource(StreamDao streamDao, StreamClientDao<Producer> producerDao
     @Path("/{streamName}")
     @Consumes(MediaType.APPLICATION_JSON)
     @Timed
-    public Response upsertStream(@ApiParam(value = "stream entity", required = true) Stream stream) {
-        // TODO: streamName path param isn't used, we should validate JSON stream matches stream name
+    public Response upsertStream(@ApiParam(value = "stream entity", required = true) Stream stream,
+                                 @ApiParam(value = "stream name", required = true) @PathParam("streamName") String streamName) {
+
```
neoword: surprised we weren't doing this already!!

neoword: one ask... can we invert the parameters?
`String streamName, Stream stream`

neoword:
Also update CHANGELOG please.

neoword:
This points out why we need better (real) IT tests. This should be added to a regression test suite.

PR #73, opened 2019-01-01
Bug/delete client error fix

### LINKED ISSUE #72: DELETE consumer service removes more than one consumer
# Expected Behavior

To remove only one consumer

# Actual Behavior

It is removing more than one consumer in a sequential way

# Steps to Reproduce
```
Having a stream W created:

1. Create consumer A for stream W
2. Create consumer B for stream W
3. Create consumer C for stream W
4. Create consumer D for stream W

5. *Delete* only consumer C
```
It will delete consumers `A`, `B`, and `C` and let only consumer `D`.

dccarlos:
# stream-registry PR

Closes #72

### Added
A new method on `StreamRegistryUtils` class to check if an actor is included in a client

### Changed
Modification on `deleteProducer` and `deleteConsumer` in `dao` package


neoword:
Also, please rebase.

core/src/main/java/com/homeaway/streamingplatform/db/dao/impl/ConsumerDaoImpl.java
```diff
@@ -188,16 +187,20 @@ private void deleteConsumer(String streamName, String consumerName) {
...
-                    iter.remove();
-                    avroStream.get().setConsumers(Lists.newArrayList(iter));
-                }
-            }
+            // Obtains consumer list size before  remove consumer
+            final int consumerInitialSize = avroStream.get().getConsumers().size();
+
+            // Obtains filtered consumer list not containing the consumer we want to remove
+            List<com.homeaway.digitalplatform.streamregistry.Consumer> withoutConsumer = avroStream.get()
+                    .getConsumers()
```
cricket007: `avroStream.get().getConsumers()` can be refactored to a variable since it's used twice

core/src/main/java/com/homeaway/streamingplatform/db/dao/impl/ProducerDaoImpl.java
```diff
@@ -148,16 +147,20 @@ private void deleteProducer(String streamName, String producerName) {
...
-                    iter.remove();
-                    avroStream.get().setProducers(Lists.newArrayList(iter));
-                }
-            }
+            // Obtains producer list size before  remove consumer
+            final int producerInitialSize = avroStream.get().getProducers().size();
+
+            // Obtains filtered producer list not containing the consumer we want to remove
+            List<com.homeaway.digitalplatform.streamregistry.Producer> withoutProducer = avroStream.get()
+                    .getProducers()
```
cricket007: `avroStream.get().getProducers()` can be refactored to a variable since it's used twice

core/src/main/java/com/homeaway/streamingplatform/db/dao/impl/ConsumerDaoImpl.java
dccarlos: It's ready

core/src/main/java/com/homeaway/streamingplatform/db/dao/impl/ProducerDaoImpl.java
dccarlos: It's ready

PR #79, opened 2019-01-03
Decouple core model API from so-called core module

cricket007:
# stream-registry PR

Introduce `api` module with `model` and `exceptions` extracted from current `core` module. 

**Goal**: Extract `validator` module, which will depend on `api`

### Added
* `api` module

### Changed
* Refactored `core` to use `api` module dependency

# PR Checklist Forms

- [x] CHANGELOG.md updated
- [ ] Reviewer assigned


prabhakar1983:
@cricket007 
Server side exception classes would not be used by clients, so i would prefer not to keep them on "stream-registry-api" module

neoword:
Lets revisit this after #84 is committed.

neoword:
Resubmit PR to `vrbo-master`

PR #111, opened 2019-02-07
BIGDATA-6059: Formatted the error responses so that the developers can debug the errors easily

prabhakar1983:
# stream-registry PR

Formatted the error responses so that the developers can debug the errors easily

### Changed
* Formatted the error responses so that the developers can debug the errors easily

# PR Checklist Forms

- [x] CHANGELOG.md updated
- [x] Reviewer assigned

### DIFF
--- a/core/src/main/java/com/homeaway/streamplatform/streamregistry/db/dao/impl/StreamDaoImpl.java
+++ b/core/src/main/java/com/homeaway/streamplatform/streamregistry/db/dao/impl/StreamDaoImpl.java
@@ -220,6 +220,8 @@ public class StreamDaoImpl extends AbstractDao implements StreamDao, StreamValid
                     upsertTopics(stream, vpc, isNewStream);
                 }
             }
+        } catch (StreamCreationException e) {
+            throw e;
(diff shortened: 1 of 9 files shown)


core/src/main/java/com/homeaway/streamplatform/streamregistry/db/dao/impl/StreamDaoImpl.java
```diff
@@ -126,61 +127,56 @@ public void upsertStream(Stream stream) {
             log.error("Stream '{}' is not valid", stream.getName());
         }
```
prabhakar1983: no changes here except, the try catch is removed. SchemaManagerException, InvalidStreamException, StreamCreationException are not handled here, rather thown back to Resource class

neoword: Yeah. I would like to schedule a pair-programming with you and @vinayakponangi and redo this with pattern moving forward. Please create an issue to cover this.

PR #124, opened 2019-02-12
Hardening dependency injection to ensure services start and stop in order

saketanisha:
# stream-registry PR

Re-organized the bean DI ordering to ensure predictability in service start and stop ordering.

### Added
* Created `StreamRegistryManagedContainer` which will be responsible for ordering lifecycle for infra components

### Changed
* Re-ordered bean instantiation in `StreamRegistryApplication.run` and registering health check post initialization of the application has finished, server is started and application is ready to handle requests now.
* Moved `KafkaManager` local to `StreamDaoImpl`
* Creating `streamRegHealthCheckStream`, `producerResource` and `consumerResource` inside `StreamRegistryHealthCheck` constructor
* Changed scope for `BaseResourceIT.createTopic` to private

# PR Checklist Forms

- [X] CHANGELOG.md updated
- [x] Reviewer assigned

### DIFF
--- a/core/src/main/java/com/homeaway/streamplatform/streamregistry/StreamRegistryApplication.java
+++ b/core/src/main/java/com/homeaway/streamplatform/streamregistry/StreamRegistryApplication.java
@@ -85,6 +85,17 @@ public class StreamRegistryApplication extends Application<StreamRegistryConfigu
 
     private MetricRegistry metricRegistry;
 
+    private InfraManager infraManager;
+
+    private RegionDao regionDao;
+    private StreamDao streamDao;
+    private StreamClientDao<Producer> producerDao;
+    private StreamClientDao<Consumer> consumerDao;
+
+    private StreamResource streamResource;
+
+    private StreamRegistryHealthCheck streamRegistryHealthCheck;
+
     public static void main(final String[] args) throws Exception {
         new StreamRegistryApplication().run(args);
     }
@@ -114,80 +125,50 @@ public class StreamRegistryApplication extends Application<StreamRegistryConfigu
 
     @Override
     public void run(final StreamRegistryConfiguration configuration, final Environment environment) {
-        Properties kstreamsProperties = new Properties();
-        configuration.getKafkaStreamsConfig().getKstreamsProperties().forEach(kstreamsProperties::put);
-        kstreamsProperties.put(ROCKSDB_CONFIG_SETTER_CLASS_CONFIG, CustomRocksDBConfig.class);
-        TopicsConfig topicsConfig = configuration.getTopicsConfig();
+        /*
+         * Step 1 - initialize managed beans
+         */
 
-        ManagedKStreams managedKStreams = new ManagedKStreams(kstreamsProperties, topicsConfig);
+        ManagedKStreams managedKStreams = createManagedKStreams(configuration);
 
-        InfraManagerConfig infraManagerConfig = configuration.getInfraManagerConfig();
-        String infraManagerClassName = infraManagerConfig.getClassName();
+        ManagedInfraManager managedInfraManager = createManagedInfraManager(configuration);
+        Preconditions.checkState(managedInfraManager != null, "managedInfraManager cannot be null.");
 
-        InfraManager infraManager = null;
-        if (infraManagerClassName != null && !infraManagerClassName.isEmpty()) {
-            try {
-                infraManager = Utils.newInstance(infraManagerClassName, InfraManager.class);
-                infraManager.configure(infraManagerConfig.getConfig());
-            } catch (Exception e) {
-                log.error("Error loading/configuring Infra Manager Class", e);
-                throw new IllegalStateException("Could not load/configure Infra Manager class", e);
-            }
-        }
-        ManagedInfraManager managedInfraManager = new ManagedInfraManager(infraManager);
+        ManagedKafkaProducer managedKafkaProducer = createManagedKafkaProducer(configuration);
 
+        /*
+         * Step 2 - initialize the DAO's
+         */
 
-        Properties producerProperties = new Properties();
-        producerProperties.put(ProducerConfig.KEY_SERIALIZER_CLASS_CONFIG, KafkaAvroSerializer.class);
-        producerProperties.put(ProducerConfig.VALUE_SERIALIZER_CLASS_CONFIG, KafkaAvroSerializer.class);
+        initDao(configuration, environment, managedKStreams, managedKafkaProducer);
 
-        configuration.getKafkaProducerConfig().getKafkaProducerProperties().forEach(producerProperties::put);
+        /*
+         * Step 3 - initialize and register the SR resources to JerseyEnvironment
+         */
 
-        ManagedKafkaProducer managedKafkaProducer = new ManagedKafkaProducer(producerProperties, topicsConfig);
+        initAndRegisterResource(environment);
 
-        final Client httpClient = new JerseyClientBuilder(environment)
-                .using(configuration.getHttpClient())
-                .using(environment)
-                .build("remoteStateStoreHttpClient");
-
-        log.info("Connection Timeout:{}", configuration.getHttpClient().getConnectionTimeout());
-        log.info("Connection Request Timeout:{}", configuration.getHttpClient().getConnectionRequestTimeout());
-
-        String env = configuration.getEnv();
-        RegionDao regionDao = new RegionDaoImpl(env, infraManager);
-
-        StreamValidator streamValidator = loadValidator(configuration, httpClient, regionDao);
-        Preconditions.checkState(streamValidator != null, "streamValidator cannot be null.");
-
-        SchemaManager schemaManager = loadSchemaManager(configuration);
-
-        KafkaManager kafkaManager = new KafkaManagerImpl();
-        StreamDao streamDao = new StreamDaoImpl(managedKafkaProducer, managedKStreams, env, regionDao, infraManager, streamValidator, schemaManager, kafkaManager);
-        StreamClientDao<Producer> producerDao = new ProducerDaoImpl(managedKafkaProducer, managedKStreams, env, regionDao, infraManager);
-        StreamClientDao<Consumer> consumerDao = new ConsumerDaoImpl(managedKafkaProducer, managedKStreams, env, regionDao, infraManager);
-
-        StreamResource streamResource = new StreamResource(streamDao, producerDao, consumerDao);
-        RegionResource regionResource = new RegionResource(regionDao);
-
-        environment.jersey().register(streamResource);
-        environment.jersey().register(regionResource);
-
-        StreamRegistryHealthCheck streamRegistryHealthCheck = new StreamRegistryHealthCheck(managedKStreams, streamResource, metricRegistry, configuration.getHealthCheckStreamConfig());
+        /*
+         * Step 4 - initialize and register monitoring and health check hooks
+         */
 
         environment.getApplicationContext().addServlet(PingServlet.class, "/ping");
-        environment.healthChecks().register("streamRegistryHealthCheck", streamRegistryHealthCheck);
+        initAndRegisterHealthCheck(configuration, environment, managedKStreams);
 
-        StreamRegistryManagedContainer managedContainer = new StreamRegistryManagedContainer(managedKStreams, managedInfraManager, managedKafkaProducer, streamRegistryHealthCheck);
-        environment.lifecycle().manage(managedContainer);
+        /*
+         * Step 5 - centralize initialization and configuration of SR server Object Mapper's
+         */
+        registerServiceMapper(environment);
+
+        /*
+         * Step 6 - In order to avoid muddle up with HK2 and Jersey lifecycle dependency we wrap up
+         * all managed beans in a centralized container to manage server components start and
+         * stop ordering
+         */
+
+        registerManagedContainer(environment, managedKStreams, managedInfraManager, managedKafkaProducer);
 
         // TODO: Make project completely based on unit tests (integration should be a separate project) (#100)
-
-        environment.getObjectMapper().setSerializationInclusion(JsonInclude.Include.ALWAYS);
-        environment.getObjectMapper()
-                .disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS);
-        GuavaModule guavaModule = new GuavaModule();
-        guavaModule.configureAbsentsAsNulls(true);
-        environment.getObjectMapper().registerModule(new GuavaModule().configureAbsentsAsNulls(true));
     }
 
     public static StreamValidator loadValidator(StreamRegistryConfiguration configuration,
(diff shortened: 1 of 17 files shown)


core/src/main/java/com/homeaway/streamplatform/streamregistry/db/dao/impl/ProducerDaoImpl.java
```diff
@@ -48,13 +51,21 @@
...
+    }
+
     public ProducerDaoImpl(ManagedKafkaProducer managedKafkaProducer,
         ManagedKStreams kStreams,
         String env,
         RegionDao regionDao,
-        InfraManager infraManager,
-        KafkaManager kafkaManager) {
-        super(managedKafkaProducer, kStreams, env, regionDao, infraManager, kafkaManager);
+        InfraManager infraManager) {
```
neoword: If this constructor is not used.... delete it... and defer the `managedInfraManager.getInfraManager()` to internal field.

core/src/main/java/com/homeaway/streamplatform/streamregistry/db/dao/impl/StreamDaoImpl.java
```diff
@@ -58,18 +61,41 @@
...
     public StreamDaoImpl(ManagedKafkaProducer managedKafkaProducer,
                          ManagedKStreams kStreams,
                          String env,
                          RegionDao regionDao,
                          InfraManager infraManager,
-                         KafkaManager kafkaManager,
                          StreamValidator validator,
                          SchemaManager schemaManager) {
-        super(managedKafkaProducer, kStreams, env, regionDao, infraManager, kafkaManager);
+        this(managedKafkaProducer, kStreams, env, regionDao, infraManager, validator, schemaManager, new KafkaManagerImpl());
```
neoword: Remove unnecessary public ctor if not needed.

core/src/main/java/com/homeaway/streamplatform/streamregistry/StreamRegistryApplication.java
```diff
@@ -174,18 +174,56 @@ public void run(final StreamRegistryConfiguration configuration, final Environme
...
+
+            @Override
+            public void onEvent(ApplicationEvent event) {
+                if (event.getType() == ApplicationEvent.Type.INITIALIZATION_FINISHED) {
+                    ServiceLocator serviceLocator =
+                        ((ServletContainer) environment.getJerseyServletContainer()).getApplicationHandler().getServiceLocator();
+
+                    StreamRegistryHealthCheck streamRegistryHealthCheck = serviceLocator.getService(StreamRegistryHealthCheck.class);
+
+                    environment.healthChecks().register("streamRegistryHealthCheck", streamRegistryHealthCheck);
```
neoword: AFAIK (might be wrong here) jersey expects its registration to happen during startup of the Application.  By deferring its registration in the event handler, are we creating possible timing issues ?

Suggestion: Move the registration inline. And make StreamRegistryHealthCheck stateful (e.g. it will be uninitialized).  In this event handler, change the state to be INITIALIZED.  If someone pings StreamRegistryHealthCheck while initializing, they will get a DOWN answer until it is INITIALIZED and passes its checks).

This may cause issues with the IT tests... but that's fine... we'll add something that waits appropriately in the IT setup() method to wait for healthcheck to be healthy.

core/src/main/java/com/homeaway/streamplatform/streamregistry/db/dao/impl/ProducerDaoImpl.java
saketanisha: Ideally we should pivot around `ManagedInfraManager` but seems existing IT impl leverages `InfraManager` and calls this constructor ( [here](https://github.com/homeaway/stream-registry/blob/master/core/src/test/java/com/homeaway/streamplatform/streamregistry/resource/BaseResourceIT.java#L248-L249) )

core/src/main/java/com/homeaway/streamplatform/streamregistry/db/dao/impl/StreamDaoImpl.java
saketanisha: Same as your previous observation - its been called from the IT flow ( [here](https://github.com/homeaway/stream-registry/blob/master/core/src/test/java/com/homeaway/streamplatform/streamregistry/resource/BaseResourceIT.java#L246-L247) )

( Maybe we've an opportunity here to streamline the IT implementation here. Will try to see if we can do it in this PR or shall we have a follow up )

saketanisha: Same as your previous observation - its been called from the IT flow ( [here](https://github.com/homeaway/stream-registry/blob/master/core/src/test/java/com/homeaway/streamplatform/streamregistry/resource/BaseResourceIT.java#L246-L247) )

( Maybe we've an opportunity here to streamline the IT implementation here. Will try to see if we can do it in this PR or shall we have a follow up )

saketanisha: ( also we've moved the KafkaManager inside `StreamDaoImpl` )

core/src/main/java/com/homeaway/streamplatform/streamregistry/StreamRegistryApplication.java
saketanisha: Came across [this](https://github.com/dropwizard/dropwizard/issues/862) moments back. Maybe will explore options sans HK2 ( inline with your feedback ).

Also will check how existing PR plays with `INITIALIZED` ( last I tested + debugged seems the service locator is instantiated post the servlet transition to `INITIALIZATION_FINISHED` state. i'll re-verify this and come back )

saketanisha: nit - confirming that during `INITIALIZATION_START` & `INITIALIZATION_APP_FINISHED` - `((ServletContainer) environment.getJerseyServletContainer()).getApplicationHandler()` is `NULL` and thus we cannot get back `ServiceLocator`.

saketanisha: [Here](https://gist.github.com/saketanisha/9406dd4b43b73e7ef00809931d84745d) is the trimmed down server startup logs.

For this experiment I removed the code inside `ApplicationEventListener` to register the health check - i.e. `environment.healthChecks().register("streamRegistryHealthCheck", streamRegistryHealthCheck);` and added a nit instrumentation to `INITIALIZATION_FINISHED` callback.

1. `/private` is registered [here](https://gist.github.com/saketanisha/9406dd4b43b73e7ef00809931d84745d#file-stream-registry-startup-log-L7).
2. `INITIALIZATION_FINISHED` callback is called [here](https://gist.github.com/saketanisha/9406dd4b43b73e7ef00809931d84745d#file-stream-registry-startup-log-L34).
3. `/private` servlet context is started [here](https://gist.github.com/saketanisha/9406dd4b43b73e7ef00809931d84745d#file-stream-registry-startup-log-L71) and before this it DW setup create `AdminEnvironment` for `/private` - see [here](https://gist.github.com/saketanisha/9406dd4b43b73e7ef00809931d84745d#file-stream-registry-startup-log-L62-L70) .
4. Finally the app http connector is bound to port 8080 and server is ready to server - [here](https://gist.github.com/saketanisha/9406dd4b43b73e7ef00809931d84745d#file-stream-registry-startup-log-L72-L73). 

I believe this change is inline with the servlet lifecycle. An observation is that its kind-of hooked with the `init` lifecycle method, i.e. semantic working of server is tightly couple with underlying servlet container.

As an experiment let me try out taking out the call back completely out of picture and drive this orchestration via single managed bean within DW that encapsulates the other building block of the system.

**_Note_**: please ignore [these](https://gist.github.com/saketanisha/9406dd4b43b73e7ef00809931d84745d#file-stream-registry-startup-log-L45-L54) error logs - the local kafka cluster not configured with infra bits.

saketanisha: ( forgot to mention above - during INITIALIZATION_START & `INITIALIZATION_APP_FINISHED` - `((ServletContainer) environment.getJerseyServletContainer()).getApplicationHandler()` is `NULL` )

saketanisha: Removed dependency on HK2 and Jersey Container lifecycle.

core/src/main/java/com/homeaway/streamplatform/streamregistry/db/dao/impl/StreamDaoImpl.java
saketanisha: Changed this flow.

core/src/main/java/com/homeaway/streamplatform/streamregistry/db/dao/impl/ProducerDaoImpl.java
saketanisha: Changed this flow.

core/src/main/java/com/homeaway/streamplatform/streamregistry/health/StreamRegistryHealthCheck.java
```diff
@@ -341,4 +338,28 @@ public String getName() {
...
+    public void start() throws Exception {
+        metricRegistry.register(Metrics.STREAM_CREATION_HEALTH.getName(), (Gauge<Integer>)() -> isStreamCreationHealthy() ? 1 : 2);
+        metricRegistry.register(Metrics.STATE_STORE_HEALTH.getName(), (Gauge<Integer>)() -> isStateStoreHealthy() ? 1 : 2);
+        metricRegistry.register(Metrics.STATE_STORE_STATE_HEALTH.getName(), (Gauge<Integer>)() -> isKStreamInValidState() ? 1 : 2);
+        metricRegistry.register(Metrics.PRODUCER_REGISTRATION_HEALTH.getName(), (Gauge<Integer>)() -> isProducerRegistrationHealthy() ? 1 : 2);
+        metricRegistry.register(Metrics.CONSUMER_REGISTRATION_HEALTH.getName(), (Gauge<Integer>)() -> isConsumerRegistrationHealthy() ? 1 : 2);
+        metricRegistry.register(Metrics.STATE_STORE_STATE.getName(), (Gauge<String>)() -> getKstreamsState().toString());
+
+        streamRegHealthCheckStream = createCanaryStream();
+        streamResource.upsertStream(streamName, streamRegHealthCheckStream);
```
neoword: isn't the point of the health check to upsert each time?

core/src/main/java/com/homeaway/streamplatform/streamregistry/health/StreamRegistryHealthCheck.java
```diff
@@ -341,4 +338,28 @@ public String getName() {
...
     }
 
+    @Override
+    public void start() throws Exception {
+        metricRegistry.register(Metrics.STREAM_CREATION_HEALTH.getName(), (Gauge<Integer>)() -> isStreamCreationHealthy() ? 1 : 2);
+        metricRegistry.register(Metrics.STATE_STORE_HEALTH.getName(), (Gauge<Integer>)() -> isStateStoreHealthy() ? 1 : 2);
+        metricRegistry.register(Metrics.STATE_STORE_STATE_HEALTH.getName(), (Gauge<Integer>)() -> isKStreamInValidState() ? 1 : 2);
+        metricRegistry.register(Metrics.PRODUCER_REGISTRATION_HEALTH.getName(), (Gauge<Integer>)() -> isProducerRegistrationHealthy() ? 1 : 2);
+        metricRegistry.register(Metrics.CONSUMER_REGISTRATION_HEALTH.getName(), (Gauge<Integer>)() -> isConsumerRegistrationHealthy() ? 1 : 2);
+        metricRegistry.register(Metrics.STATE_STORE_STATE.getName(), (Gauge<String>)() -> getKstreamsState().toString());
```
neoword: not sure why you decided to this on start() instead of in the constructor.
Please leave a comment as to why this is , or revert this back.

core/src/main/java/com/homeaway/streamplatform/streamregistry/health/StreamRegistryHealthCheck.java
neoword: Please leave comment as to why you have this section in `start()` instead of during constructor time.

core/src/main/java/com/homeaway/streamplatform/streamregistry/health/StreamRegistryHealthCheck.java
saketanisha: Will revert this change.

core/src/main/java/com/homeaway/streamplatform/streamregistry/health/StreamRegistryHealthCheck.java
saketanisha: Will revert this change.

core/src/main/java/com/homeaway/streamplatform/streamregistry/StreamRegistryApplication.java
```diff
@@ -114,87 +125,50 @@ protected SwaggerBundleConfiguration getSwaggerBundleConfiguration(StreamRegistr
...
-                infraManager.configure(infraManagerConfig.getConfig());
-            } catch (Exception e) {
-                log.error("Error loading/configuring Infra Manager Class", e);
-                throw new IllegalStateException("Could not load/configure Infra Manager class", e);
-            }
-        }
-        ManagedInfraManager managedInfraManager = new ManagedInfraManager(infraManager);
+        /*
+         * Step 2 - initialize the DAO's
+         */
```
saketanisha: **_Note:_** - While re-factoring `createManagedKStreams`, `createManagedInfraManager` and `createManagedKafkaProducer` sensed an opportunity to centralize DAO creation in a separate method. Should this looks to be an over kill then open to make Step `2` inline and also for the other steps below.

core/src/main/java/com/homeaway/streamplatform/streamregistry/StreamRegistryApplication.java
```diff
@@ -264,4 +225,124 @@ public void setConfig(final String storeName, final Options options, final Map<S
...
+        ManagedKStreams managedKStreams = new ManagedKStreams(kstreamsProperties, topicsConfig);
+        return managedKStreams;
+
+    }
+
+    private ManagedInfraManager createManagedInfraManager(final StreamRegistryConfiguration configuration) {
+        InfraManagerConfig infraManagerConfig = configuration.getInfraManagerConfig();
+        String infraManagerClassName = infraManagerConfig.getClassName();
+
+        if (infraManagerClassName != null && !infraManagerClassName.isEmpty()) {
```
neoword: should we require an infra manager ?
(e.g. should there be a precondition check with the opposite of this?)

saketanisha: Added the pre-condition to ensure that `infraManager` is null before the `ManagedInfraManager` is created.

saketanisha: Ok got it. Maybe `null` check on `infraManagerClassName` as `InfraManagerConfig.className` is marked `@NotNull` ( [here](https://github.com/homeaway/stream-registry/blob/master/core/src/main/java/com/homeaway/streamplatform/streamregistry/configuration/InfraManagerConfig.java#L31-L36) )

I'll update with `isEmpty` Preconditions check.

PR #127, opened 2019-02-18
added helm chart for deploying stream registry

ayansen:
Added a helm chart to enable easy deployment of stream registry in a kubernetes cluster.

### DIFF
--- a/charts/README.md
+++ b/charts/README.md
@@ -1,0 +1,47 @@
+# Stream-Registry Helm Chart
+This helm chart creates a [Homeaway Stream-Registry server](https://homeaway.github.io/stream-registry/).
+
+## Prerequisites
+* Kubernetes 1.6
+* A running Kafka Installation
+* A running Zookeeper Installation
+* A running confluent schema registry
+
+## Chart Details
+
+This chart will do the following:
+
+* Deploy Stream registry and all its dependencies
+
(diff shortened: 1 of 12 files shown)


charts/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+This chart will do the following:
+
+* Deploy Stream registry and all its dependencies
+
+## Installing the Chart
+
+To install the chart, use the following:
+1. clone this github repo code
+2. configure helm to point to kubernetes cluster (for minikube refer [here](https://docs.helm.sh/using_helm/) )
```
neoword: do you also need to setup helm incubator repos?
- e.g. `helm repo add incubator https://kubernetes-charts-incubator.storage.googleapis.com/` ??

Also do you need to setup a local http helm repo?

e.g.
`helm serve --repo-path ./charts`


finally... do you need to add a .gitignore for these files?
- charts/charts/
- charts/requirements.lock

??

neoword:
image
 ❤️

charts/README.md
ayansen: yes, we need to add the helm incubator repo. And we should also update the gitignore. 
The local repo might not be needed since we are passing the path to the chart during install.

ayansen: updated the pr.

PR #137, opened 2019-03-07
Append status code to log statement

prabhakar1983:
# stream-registry PR

Earlier Log statements used to have http response code, which users used to search Splunk using response code.
So, Appended status code to log statement.

### Added
Append Http Response Status Code to the log statement

# PR Checklist Forms

- [x] Reviewer assigned


core/src/main/java/com/homeaway/streamplatform/streamregistry/resource/BaseResource.java
```diff
@@ -25,12 +25,13 @@
...
+                e.getCause() != null ? e.getCause().getMessage() : null);
 
+        log.error(errorMessage.toString(), e);
         return Response.status(status)
-                .entity(new ErrorMessage(status.getStatusCode(),
-                        e.getMessage(),
-                        e.getCause() != null ? e.getCause().getMessage() : null))
+                .entity(errorMessage)
                 .build();
     }
```
prabhakar1983: Earlier, the log message did not print Response Code.

neoword: Sounds great. Do you think unit test is necessary so that we don't get a regression ?

PR #139, opened 2019-03-11
Updating swagger example values

arunvasudevan:
# stream-registry PR

Changing the swagger example values

### Changed
- Changed the example values for swagger

# PR Checklist Forms

- [x] CHANGELOG.md updated
- [x] Reviewer assigned

### DIFF
--- a/core/src/main/java/com/homeaway/streamplatform/streamregistry/model/Stream.java
+++ b/core/src/main/java/com/homeaway/streamplatform/streamregistry/model/Stream.java
@@ -27,6 +27,8 @@ import lombok.Data;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
 
+import io.swagger.annotations.ApiModelProperty;
+
 import com.homeaway.digitalplatform.streamregistry.SchemaCompatibility;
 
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/homeaway/streamplatform/streamregistry/model/Stream.java
```diff
@@ -121,11 +124,13 @@
     /**
      * Number of Partitions
      */
+    @ApiModelProperty(example = "1")
     int partitions;
 
     /**
      * Replication Factor
```
prabhakar1983: Will the replicationFactor reflect as 3 in the Swagger ?

If so, then it may fail, because, usually OSS testing in laptop will have only 1 Broker clusters

arunvasudevan: @prabhakar1983 as discussed setting replicationFactor as 1 by default would be a wrong standard in production. In local they might have to update the replicationFactor to 1

neoword: agreed that we should opt for production, and maybe provide a readme or documentation for local developer values.

prabhakar1983: 👍

Issue #146, opened 2019-03-26
Get clusters throws 500


arunvasudevan:
# Expected Behavior

When searching for invalid values in search field it should throw not found 

# Actual Behavior

It throws a 500 error

# Steps to Reproduce

1. Get Clusters
2. Search for "invalid" in the type field.
3. Throws a 500 error.

scunniff:
Should this be closed? Based on Arun's PR (#147) it looks like this was fixed.

PR #147, opened 2019-03-26
Handling NPE for getClusters

arunvasudevan:
# stream-registry PR

Fixes Null Pointer Exception in getClusters

### Changed
- Added Null check in the filtering of the results to handle null pointer exception.
- Fixes bug #146 

# PR Checklist Forms

- [X] CHANGELOG.md updated
- [X] Reviewer assigned
- [X] PR assigned (presumably to submitter)
- [X] Labels added (enhancement, bug, documentation)

### DIFF
--- a/core/src/main/java/com/homeaway/streamplatform/streamregistry/resource/ClusterResource.java
+++ b/core/src/main/java/com/homeaway/streamplatform/streamregistry/resource/ClusterResource.java
@@ -17,6 +17,7 @@ package com.homeaway.streamplatform.streamregistry.resource;
 
 import java.util.List;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import javax.ws.rs.Consumes;
 import javax.ws.rs.GET;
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/homeaway/streamplatform/streamregistry/resource/ClusterResource.java
```diff
@@ -145,10 +145,14 @@ public Response getAllClusters(@ApiParam(value = "clusterName") @QueryParam("clu
...
             if(type != null && !type.isEmpty()) {
                 clusterList = clusterList.stream()
+                    .filter(e -> e.getClusterKey().getType()!= null)
                     .filter(e -> e.getClusterKey().getType().equalsIgnoreCase(type))
                     .collect(Collectors.toList());
             }
-
+            if (clusterList.isEmpty()) {
+                return Response
+                    .status(Response.Status.NOT_FOUND.getStatusCode()).build();
```
cricket007: I think a `200 OK []` response is fine. 

> if I ask for a set of objects that match a query, a null set is a valid response and I want that returned with a 200 code. The rationale for this is that the query was valid, it succeeded and the query returned nothing.

https://softwareengineering.stackexchange.com/a/203493/279825

cricket007: Plus, is there really a need to continuously collect the stream to a list rather than continuing to filter it just as a `Stream` object?

PR #158, opened 2019-04-15
Add log statements when a Stream's topic gets created/updated.

prabhakar1983:
# stream-registry PR

Added log statements when a topic is created so that alerting can be added for state changes.

# PR Checklist Forms

- [ ] CHANGELOG.md updated
- [ ] Reviewer assigned


core/src/main/java/com/homeaway/streamplatform/streamregistry/service/impl/StreamServiceImpl.java
```diff
@@ -216,6 +216,8 @@ public void deleteStream(String streamName) throws StreamNotFoundException{
             throw new StreamNotFoundException(String.format("Stream=%s not found.", streamName));
         }
 
+        // TODO: issue#156. Delete the underlying Topic when a Stream is deleted.
+        // It is dangerous to delete the topic in this case without enough authorization, monitoring and alerting in place
```
neoword: Suggest that DELETE be moved to a queue for manual deletion for now.

When we have sufficient RBAC in place, this can be done "on command".

Issue #35, opened 2018-12-06
Stream /replications are supported


neoword:
Desired Behavior
==============================

`/replications` become a first class citizen of stream registry

Benefits
==============================

* Enables easy replication of streams across clusters or namespaces
* Supports intra-namespace replication
* Supports inter-namespace replication

PR #169, opened 2019-04-24
Removing creation of topics in replicated vpcs

krishnatg:
# stream-registry PR

Topics are being created in the VPCs where streams are to be replicated, with no added benefit of having created them in the first place. Stream Replicator application is responsible for creating topics in those VPCs and Stream Registry should not do anything about them.

### Deleted
* The creation of topics in the VPCs where the stream is to be replicated

# PR Checklist Forms

- [x] CHANGELOG.md updated
- [x] Reviewer assigned

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [0.5.1] - SNAPSHOT
 ### Added
 - Updated `README` with Expedia Group stream registry announcement 
+- Removed the creation of topics in the VPCs where it is to be replicated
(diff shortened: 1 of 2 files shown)


neoword:
See #35 for context.

Confused.  Shouldn't stream registry _**be exactly responsible**_ for stream-binding creation (e.g. topics) in a target namespace?

This feels like functionality that we would only add later.

Stream Replicator agent _**should NOT be responsible**_ for stream-binding creation.  If the stream-binding does NOT exist, it should fail.

See #35 for context.

neoword:
Marking `wont do`

krishnatg:
Need to re-open this

PR #52, opened 2018-12-12
Stop Stream Creation for existing topics if configs don't match

arunvasudevan:
# stream-registry PR

Check if the topic configs as part of the new stream are same as that of the existing topics config. If the configs don't match throw Stream Creation Exception.

## Changed
- Stop Stream Creation for Existing topics if topic configs don't match

## Deleted

# PR Checklist Forms

- [X] This template filled (above this section)
- [X] CHANGELOG.md updated
- [x] Reviewer assigned
- [x] PR assigned (presumably to submitter)
- [x] Labels added (enhancement, bug, documentation) 
- [x] Milestone selected
- [X] All files reviewed before sending to reviewers
- [X] PR builds successfully

PR #170, opened 2019-04-26
adding the forceSync flag for handling conflicting topic configs

### LINKED ISSUE #114: Consider adding forceSync=true for topic updates
# Desired Behavior

Add `forceSync=true` for updating of topic metadata.

# Benefits

* See https://github.com/homeaway/stream-registry/blob/bacfe0a5d61fa4102bd0ff521f80d60ff90257e2/core/src/main/java/com/homeaway/streamplatform/streamregistry/db/dao/impl/KafkaManagerImpl.java#L128-L134
* By adding `forceSync=true` , when users get the error, they can use the `forceSync=true` to get the value from the underlying broker, so that they can then update.

krishnatg:
# stream-registry PR

Adding flag to provide an option to override user-provided topic config in case of already existing conflicting configs on the underlying brokers (https://github.com/ExpediaGroup/stream-registry/issues/114)

- Fixes #114 

### Added
* Boolean flag `forceSync` which can be set by a user while upserting a stream to override the provided topic configs in the stream and continue using the existing topic configs on the brokers.

# PR Checklist Forms

- [x] CHANGELOG.md updated
- [x] Reviewer assigned

### DIFF
--- a/infra-provider-kafka/src/main/java/com/homeaway/streamplatform/streamregistry/provider/impl/KafkaInfraManager.java
+++ b/infra-provider-kafka/src/main/java/com/homeaway/streamplatform/streamregistry/provider/impl/KafkaInfraManager.java
@@ -180,7 +180,7 @@ public class KafkaInfraManager implements InfraManager {
     }
 
     @Override
-    public void upsertTopics(Collection<String> topics, int partitions, int replicationFactor, Properties properties, boolean isNewStream)
+    public void upsertTopics(Collection<String> topics, int partitions, int replicationFactor, Properties properties, boolean isNewStream, boolean forceSync)
         throws StreamCreationException {
         // TODO - Cannot guarantee against race conditions... should probably move to event-source paradigm to
         //      protect against this (and maybe employ optimistic locking for extra safety).
@@ -202,7 +202,7 @@ public class KafkaInfraManager implements InfraManager {
             List<String> topicsToCreate = partitionMaps.get(false);
 
             // update any topics that are necessary
-            updateTopics(zkUtils, topicsToUpdate, topicConfigMap, isNewStream);
+            updateTopics(zkUtils, topicsToUpdate, topicConfigMap, isNewStream, forceSync);
 
             // now create any topics that were necessary to create this run
             createTopics(zkUtils, topicsToCreate, partitions, replicationFactor, topicConfigMap);
@@ -233,7 +233,7 @@ public class KafkaInfraManager implements InfraManager {
         return topicExists;
     }
 
-    void updateTopics(ZkUtils zkUtils, List<String> topicsToUpdate, Map<String, String> topicConfigMap, boolean isStreamNotAvailableInStreamRegistryDB)
+    void updateTopics(ZkUtils zkUtils, List<String> topicsToUpdate, Map<String, String> topicConfigMap, boolean isStreamNotAvailableInStreamRegistryDB, boolean forceSync)
         throws StreamCreationException {
         for (String topic : topicsToUpdate) {
             // update topic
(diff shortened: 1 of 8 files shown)


infra-provider-kafka/src/main/java/com/homeaway/streamplatform/streamregistry/provider/impl/KafkaInfraManager.java
```diff
@@ -254,8 +254,12 @@ void updateTopics(ZkUtils zkUtils, List<String> topicsToUpdate, Map<String, Stri
             // to exactly match downstream config when the stream-registry has not "onboarded" existing topic
             // for the first time.
 
-            // TODO Alternatively we can add a forceSync=true flag, ignoring any user provided info, and only updating SR with the underlying settings
-            //      We should probably do forceSync=true anyway, as it provides a simple way to keep things in sync (#114)
+            // Added a forceSync=true flag, ignoring any user provided info, and only updating SR with the underlying settings (#114)
+            if (forceSync) {
+                // NOTHING TO DO!
+                log.info("topic config mismatch for {} ignored. Input config={}, actual (retained) config={}", topic, topicConfigMap, actualTopicConfig);
+                continue;
```
cricket007: What exactly is being "forced" if all you do is `continue`?

krishnatg: Well, the flag is "forcing" the stream-registry to sync the given topic configs with what exists already on the underlying brokers.
I guess I can say that by `continue`ing here, the code is "forcing" the flow to not go ahead to the following exception generation ..

cricket007: I put alternative names in #114

cricket007: > ignoring any user provided info, and only updating SR with the underlying settings

The way I interpret that is to take `actualTopicConfig` and overwrite `topicConfigMap` with it...

neoword:
Some clarification on `forceSync=true`:
- This should SUCCEED, IF AND ONLY IF config is THE SAME on all underlying clusters
- This should FAIL, if any of the cluster (stream-bindings) differ, with an error having a call to action to make fix the corresponding clusters that are different.

neoword:
Waiting for more info from team before I comment on this PR.

infra-provider-kafka/src/main/java/com/homeaway/streamplatform/streamregistry/provider/impl/KafkaInfraManager.java
neoword: Doesn't `continue` actually perform the UPDATE???
I agree with @cricket007 remarks on `actualTopicConfig` and `topicConfigMap`... where we should be overriding what the user provided, with what is _actually_ out there.

neoword:
Also should `forceSync=true` fail if no underlying topic exists?

infra-provider-kafka/src/main/java/com/homeaway/streamplatform/streamregistry/provider/impl/KafkaInfraManager.java
cricket007: > Doesn't continue actually perform the UPDATE?

No, it breaks here and goes to the next iteration of the loop. `updateTopic` isn't called.

neoword: Oh. I missed a curly-brace close.

line 270 is INSIDE not... outside the loop.
https://github.com/ExpediaGroup/stream-registry/pull/170/files/09090b33c68b43bf046a36ddd7b7ed2959b849e3#diff-b4dcb632004addc439896ec21e5b7dabR270

arunvasudevan:
Here are my comments on the PR:
1. `forceSync` is definitely a mis-guiding name..either `ignoreTopicCreationIfConfigsMismatch` or `skipTopicCreationIfConfigsMismatch` or something along that lines.

2. I can see this PR as a followup to the PR (https://github.com/ExpediaGroup/stream-registry/pull/52).
If the provided configs are different from the underlying topic config we have 2 options:
    - Force the underlying topic with the provided config  -> We don't want to do this based on the above PR (#52) and the issue that led to it.
    - Ignore (or skip) topic creation -> This is safer than the above option but the user may not notice the log in the CI/CD and it could lead to unexpected behavior later in the application.

Why not create topic only at the vpc and not in the replicatedVPC? creating topics in the replicatedVPC is a replication tools task isn't it?

krishnatg:
Since (https://github.com/ExpediaGroup/stream-registry/pull/178) is merged, this PR is not required.

PR #178, opened 2019-05-08
Removal of redundant topic creation in replicated VPCs

krishnatg:
# stream-registry PR

Topics are being created in the VPCs where streams are to be replicated, with no added benefit of having created them in the first place.

### Deleted
* The creation of topics in the VPCs where the stream is to be replicated

# PR Checklist Forms

- [x] CHANGELOG.md updated
- [x] Reviewer assigned

### DIFF
--- a/core/src/main/java/com/homeaway/streamplatform/streamregistry/service/impl/StreamServiceImpl.java
+++ b/core/src/main/java/com/homeaway/streamplatform/streamregistry/service/impl/StreamServiceImpl.java
@@ -157,7 +157,6 @@ public class StreamServiceImpl extends AbstractService implements StreamService 
      */
     private void verifyAndUpsertTopics(Stream stream, boolean isNewStream) throws StreamCreationException, ClusterNotFoundException {
         List<String> vpcList = stream.getVpcList();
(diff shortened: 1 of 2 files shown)


krishnatg:
If this looks good, can this PR be merged?

core/src/main/java/com/homeaway/streamplatform/streamregistry/service/impl/StreamServiceImpl.java
```diff
@@ -157,17 +157,10 @@ private void applyDefaultPartition(Stream stream) {
      */
     private void verifyAndUpsertTopics(Stream stream, boolean isNewStream) throws StreamCreationException, ClusterNotFoundException {
         List<String> vpcList = stream.getVpcList();
```
neoword: Might want to put in a comment here and reference the issue....
For future developers to know what's going on here... and the thinking...

Maybe something like:
```
// Only pay attention to "primary" (not replacted) vpcList that way if replicated targets are different
// we do not get "stuck". Reference the appropriate issue (#114) and (#52) that has the the context.
```

PR #187, opened 2019-07-10
Cleanup model

nahguam:
# stream-registry PR

Cleans up the model:
- Add type/configration to Domain.
- Add DomainHandler
- Entity relationships properly modelled with FKs

# PR Checklist Forms

- [ ] CHANGELOG.md updated
- [ ] Reviewer assigned


core/src/main/java/com/expediagroup/streamplatform/streamregistry/core/validator/EntityValidator.java
```diff
@@ -31,6 +35,14 @@
   public void validate(Entity<?> entity, Optional<? extends Entity<?>> existing) {
```
rshotels: I think the Optional here obscures the fact that there are two possible validation paths. I also don't think we should be using except as a parameter type.

PR #188, opened 2019-07-12
Use bull for DTO conversion

nahguam:
# stream-registry PR

Use bull for DTO conversion

# PR Checklist Forms

- [ ] CHANGELOG.md updated
- [ ] Reviewer assigned


graphql-api/src/main/java/com/expediagroup/streamplatform/streamregistry/graphql/model/GraphQLDomain.java
```diff
@@ -15,30 +15,18 @@
...
-import com.expediagroup.streamplatform.streamregistry.model.Domain;
-
 @Value
 @Builder
 public class GraphQLDomain {
   String name;
   String owner;
   String description;
-  Iterable<GraphQLKeyValue> tags;
+  List<GraphQLKeyValue> tags;
```
rshotels: Are we sure about this? Seems like tags should be a unique set.
Why the change from Iterable  to List, does BUL require this?

graphql-api/src/main/java/com/expediagroup/streamplatform/streamregistry/graphql/model/GraphQLKeyValue.java
```diff
@@ -29,23 +28,20 @@
...
-    return map
-        .entrySet()
+  public static List<GraphQLKeyValue> toMap(Map<String, String> map) {
+    return Optional
+        .ofNullable(map)
         .stream()
+        .flatMap(m -> m.entrySet().stream())
         .map(e -> new GraphQLKeyValue(e.getKey(), e.getValue()))
-        .collect(toList());
+        .collect(Collectors.toList());
```
rshotels: consider a static import

graphql-api/src/main/java/com/expediagroup/streamplatform/streamregistry/graphql/model/GraphQLDomain.java
nahguam: Yes, BUL requires the change in type. But yes you're right - it should be a Set with comparator on the key.

nahguam: We can look at that separately from this PR

graphql-api/src/main/java/com/expediagroup/streamplatform/streamregistry/graphql/model/GraphQLKeyValue.java
nahguam: ugh

PR #196, opened 2019-08-13
Scalars

nahguam:
# stream-registry PR

Introduce `Tags` and `Config` scalar type to replace `[KeyValue]`

# PR Checklist Forms

- [ ] CHANGELOG.md updated
- [ ] Reviewer assigned


core/src/main/java/com/expediagroup/streamplatform/streamregistry/core/predicate/ObjectNodePatternMatchPredicateFactory.java
```diff
@@ -0,0 +1,42 @@
...
+
+import org.springframework.stereotype.Component;
+
+@Component
+public class ObjectNodePatternMatchPredicateFactory {
+  private static final ObjectNode emptyObjectNode = new ObjectMapper().createObjectNode();
+  private final PatternMatchPredicateFactory patternMatchPredicateFactory = new PatternMatchPredicateFactory();
+
+  public <T> Predicate<T> create(T query, Function<T, ObjectNode> function) {
+    if (Optional
```
rshotels: This is what you mentioned about having a query object but leaving support as future work presumably?

core/src/test/java/com/expediagroup/streamplatform/streamregistry/core/handler/HandlerWrapperTest.java
```diff
@@ -34,11 +35,13 @@
 
 @RunWith(MockitoJUnitRunner.class)
 public class HandlerWrapperTest {
+  private static final ObjectMapper mapper = new ObjectMapper();
```
rshotels: does this need to be a field, or could we use:

.configuration(new ObjectMapper().createObjectNode())

graphql-api/src/main/resources/stream-registry.graphql
```diff
@@ -89,24 +77,24 @@ type Mutation {
   upsertDomain(
     name: String!
     description: String!
-    tags: [KeyValueInput!]!
+    tags: Tags!
     type: String!
-    configuration: [KeyValueInput!]!
+    configuration: Config!
   ): Boolean!
```
rshotels: I think these should return the outcome Domain / Schema / Stream

graphql-api/src/main/resources/stream-registry.graphql
```diff
@@ -1,49 +1,37 @@
-type KeyValue {
-  key: String!
-  value: String!
-}
+scalar Tags
+scalar Config
 
 type Domain {
   name: String!
```
rshotels: do these objects need an id as well as a name, or are they one and the same?

graphql-client-support/src/main/java/com/expediagroup/streamplatform/streamregistry/graphql/client/ConfigTypeAdapter.java
```diff
@@ -0,0 +1,50 @@
...
+import java.io.IOException;
+import java.io.UncheckedIOException;
+
+import com.apollographql.apollo.response.CustomTypeAdapter;
+import com.apollographql.apollo.response.CustomTypeValue;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+
+public class ConfigTypeAdapter implements CustomTypeAdapter<Object> {
```
rshotels: Nice

graphql-api/src/main/resources/stream-registry.graphql
nahguam: GraphQL has an `ID` type but each entity has it's own composite primary key. domain is the name, schema is the schema name and domain name, and stream is the stream name, domain name and stream version.

graphql-api/src/main/resources/stream-registry.graphql
nahguam: I think I agree.

PR #200, opened 2019-09-04
Pulling jib plugin config out

wardolph:
Pulling jib config out. Now you can change jib goal on maven package.

### DIFF
--- a/app/pom.xml
+++ b/app/pom.xml
@@ -65,20 +65,12 @@
         </configuration>
         <executions>
           <execution>
-            <id>${jib-maven-plugin.goal}</id>
+            <id>${jib.goal}</id>
             <goals>
-              <goal>${jib-maven-plugin.goal}</goal>
+              <goal>${jib.goal}</goal>
             </goals>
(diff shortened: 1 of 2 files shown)


app/pom.xml
```diff
@@ -65,9 +65,9 @@
         </configuration>
         <executions>
           <execution>
-            <id>dockerBuild</id>
```
nahguam: I presume you're doing this so you can use the `build` goal? If so, could you also remove the commented out TODO execution below this one?

pom.xml
```diff
@@ -83,6 +83,7 @@
     <maven-javadoc-plugin.version>3.1.0</maven-javadoc-plugin.version>
     <maven-compiler-plugin.version>3.8.0</maven-compiler-plugin.version>
     <jib-maven-plugin.version>1.4.0</jib-maven-plugin.version>
+    <jib-maven-plugin.goal>dockerBuild</jib-maven-plugin.goal>
```
nahguam: `jib.goal` for brevity

PR #202, opened 2019-09-04
Split all Upserts in the API to Insert and Update

rshotels:
# stream-registry PR

_&lt;Split all Upserts in the API to Insert and Update&gt;_

### Added
* _&lt; Insert and Update&gt;_

### Changed
* _&lt;The Graphql internal and Client api definitions&gt;_
* _&lt;internally insert and update still defer to the upsert methods&gt;

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 * Reference GraphQL client.
 
 ### Changed
+- Seperated Upsert into Insert and Update _at the API layer only_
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 * Reference GraphQL client.
 
 ### Changed
+- Seperated Upsert into Insert and Update _at the API layer only_
```
nahguam:
```suggestion
- Separated Upsert into Insert and Update _at the API layer only_
```

PR #206, opened 2019-09-21
Add docker-compose for running dependent services and project

tetriscode:
# stream-registry PR

Add docker-compose for development and testing deployment. It does require that the container be pre built prior to running. 

### Added
* docker-compose.yml

# PR Checklist Forms

- [x] CHANGELOG.md updated
- [ ] Reviewer assigned

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,7 +30,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Replaced ConfluentSchemaManager with simple HTTP impl.
 - Reformat 2 spaces indent
 - Removed unused dependencies & plugins
-- License update  
+- License update
+- docker-compose support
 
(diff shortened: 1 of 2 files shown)


tetriscode:
I am not able to add labels or assign a reviewer.

cricket007:
Bump? @nahguam @teabot

nahguam:
In its current form, Stream Registry does not produce a docker image. Out of the box it has only 'default' configuration types. Expedia has handlers for internal business rules which we hope to eventually open source along with other components in the platform. At that point it would make sense to provide a solution like this so a developer would able to spin up a whole platform for debugging and testing.

cricket007:
> In its current form, Stream Registry does not produce a docker image 

I added that... 

https://github.com/ExpediaGroup/stream-registry/blob/master/app/pom.xml#L48

PR #208, opened 2019-11-06
Converted ITs into suite

nahguam:
Converted ITs into suite so we only need to start postgres and the app once.

# PR Checklist Forms

- [ ] CHANGELOG.md updated
- [ ] Reviewer assigned

### DIFF
--- a/it/src/test/java/com/expediagroup/streamplatform/streamregistry/it/StreamRegistryIT.java
+++ b/it/src/test/java/com/expediagroup/streamplatform/streamregistry/it/StreamRegistryIT.java
@@ -1,0 +1,96 @@
+/**
+ * Copyright (C) 2018-2019 Expedia, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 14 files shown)


it/src/test/java/com/expediagroup/streamplatform/streamregistry/it/StreamRegistryIT.java
```diff
@@ -0,0 +1,96 @@
...
+          .withEnv("POSTGRES_PASSWORD", "streamregistry")
+          .withEnv("POSTGRES_DB", "streamregistry")
+          .withFixedExposedPort(POSTGRES_PORT, 5432);
+
+  @BeforeClass
+  public static void before() {
+    String[] args = new String[]{
+        "--server.port=0",
+        "--schema.registry.url=http://schema-registry",
+        "--spring.datasource.url=jdbc:postgresql://localhost:" + POSTGRES_PORT + "/streamregistry",
```
bsideup: FYI it is recommended to not use "localhost", but `container.getContainerIpAddress()`.
See https://www.testcontainers.org/features/networking/#getting-the-container-ip-address

Also, there is no need for `randomPort()`, you can just use `withExposedPort(5432)` and `getMappedPort(5432)`


Last but not least, you can simply use [the JDBC support](https://www.testcontainers.org/modules/databases/#database-containers-launched-via-jdbc-url-scheme), remove the container definition and use the following JDBC URL:
```
jdbc:tl:postgresql:///streamregistry
```
(note that you may need to set `spring.datasource.driver-class-name=org.testcontainers.jdbc.ContainerDatabaseDriver` too)

PR #218, opened 2019-12-09
Added StreamRegistryApolloClient.builder to create client with requir…

rshotels:
Added StreamRegistryApolloClient.builder to create client with required adapters

### DIFF
--- a/graphql-client-support/pom.xml
+++ b/graphql-client-support/pom.xml
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
   <modelVersion>4.0.0</modelVersion>
 
   <parent>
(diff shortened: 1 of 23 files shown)


graphql-client-support/pom.xml
```diff
@@ -40,6 +41,24 @@
       <artifactId>mockito-core</artifactId>
       <scope>test</scope>
     </dependency>
+
+    <dependency>
+      <groupId>com.expediagroup.streamplatform</groupId>
```
nahguam: We should not force the `stream-registry-graphql-client` and `stream-registry-graphql-client-reactor` dependencies on users who simply want the `ObjectNodeTypeAdapter` support.

rshotels: How do you propose we fix that?

graphql-client-support/src/main/java/com/expediagroup/streamplatform/streamregistry/graphql/client/builder/StreamRegistryApolloClient.java
```diff
@@ -0,0 +1,32 @@
...
+import com.apollographql.apollo.ApolloClient;
+
+import com.expediagroup.streamplatform.streamregistry.graphql.client.adapters.ObjectNodeTypeAdapter;
+import com.expediagroup.streamplatform.streamregistry.graphql.client.type.CustomType;
+
+public class StreamRegistryApolloClient {
+
+  public static ApolloClient.Builder builder() {
+    return ApolloClient.builder()
+        .okHttpClient(new OkHttpClient.Builder().build())
```
nahguam: You should allow users the option to configure the `OkHttpClient`

Maybe it would be better as a configure method?

```
public static ApolloClient configure(ApolloClient.Builder builder) {
  return builder
    .addCustomTypeAdapter(CustomType.OBJECTNODE, new ObjectNodeTypeAdapter())
    .build();
```

rshotels: They can do, by calling okHttpClient on the returned builder.

nahguam: fair enough :)

PR #226, opened 2020-02-09
feature/[STREAMP-1787]-cud notification event multicaster

dccarlos:
# stream-registry PR

General support for emitting *notification events* from the core services, in this case for _CREATE_, _UPDATE_ and _DELETE_ entities

### Added
* Added a spring application event multicaster
* Added an interface for services to multicast CUD events
* Unit tests

### Changed
* Changed all core service to support `NotificationEventEmitter`
* Changed all core service CUD methods to emit notification events

# PR Checklist Forms

- [x] CHANGELOG.md updated
- [ ] Reviewer assigned


core/src/main/java/com/expediagroup/streamplatform/streamregistry/core/events/NotificationEvent.java
```diff
@@ -0,0 +1,46 @@
...
+import lombok.NonNull;
+import lombok.ToString;
+
+import org.springframework.context.ApplicationEvent;
+import org.springframework.core.ResolvableType;
+import org.springframework.core.ResolvableTypeProvider;
+
+@Getter
+@ToString
+@EqualsAndHashCode
```
rshotels: Consider replacing these with @Data

dccarlos: I tried to set `@Data`, but it doesn't compile.
The reason is that I'm extending an abstract class `org.springframework.context.ApplicationEvent` and `@Data` can't set in a subclass, It needs to be defined in the base class (due to `@RequiredArgsConstructor`) so I let only the ones required so it can be more _inmutable_ :)

dccarlos: I just set `@Data` it compiled properly.
 
As note: It seems intellij has an error on `@Data` when it is used on subclasses showing ```lombok needs a default constructor in the base class```

dccarlos:
Based on a conversation I had with @nahguam about the convenience of the injection of a service instead the implementation of `NotificationEventEmitter` (composition over inheritance), I'll perform the proper modification 👍

core/src/main/java/com/expediagroup/streamplatform/streamregistry/core/services/ConsumerService.java
```diff
@@ -22,18 +22,25 @@
...
 import com.expediagroup.streamplatform.streamregistry.core.repositories.ConsumerRepository;
 import com.expediagroup.streamplatform.streamregistry.core.validators.ConsumerValidator;
 import com.expediagroup.streamplatform.streamregistry.model.Consumer;
 import com.expediagroup.streamplatform.streamregistry.model.keys.ConsumerKey;
 
+@Slf4j
 @Component
 @RequiredArgsConstructor
-public class ConsumerService {
+public class ConsumerService implements NotificationEventEmitter<Consumer> {
```
nahguam: We've discussed already but to capture the comment here - would make for less repetition if the `NotificationEventEmitter` were composed into the Service rather than inheriting it.

nahguam: Ah didn't see comment above. I'll resolve this thread.

PR #227, opened 2020-02-19
fix entity status inconsistency

a-nersharma:
# stream-registry PR

Consistent status for all entities

### Changed
* applied Consistent status for all entities

# PR Checklist Forms

- [x] CHANGELOG.md updated
- [x] Reviewer assigned

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [0.9.5] - 20200219
+### Bugfix
+- Status field will return blank object when it is null for all the entities.
+
 ## [0.9.4] - 20200206
 ### Added
 - Added an `authentication_group` tag to the `graphql_api` metric.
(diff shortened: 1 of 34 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [0.9.5] - 20200219
```
nahguam: Add the new entry as: `## [Unreleased]` - the version and date will be added after release

app/pom.xml
```diff
@@ -5,7 +5,7 @@
   <parent>
     <groupId>com.expediagroup.streamplatform</groupId>
     <artifactId>stream-registry</artifactId>
-    <version>0.9.5-SNAPSHOT</version>
+    <version>0.9.6-SNAPSHOT</version>
```
nahguam: No need to version bump. Leave as `0.9.5-SNAPSHOT`

model/src/main/java/com/expediagroup/streamplatform/streamregistry/model/Zone.java
```diff
@@ -30,4 +30,9 @@
   private ZoneKey key;
   private Specification specification;
   private Status status;
+
```
nahguam: Looks like we're missing a `ZoneResolver`. Perhaps create one so that the solution for the status is consistent across all entities?

CHANGELOG.md
a-nersharma: done

model/src/main/java/com/expediagroup/streamplatform/streamregistry/model/Zone.java
a-nersharma: done

app/pom.xml
a-nersharma: done

PR #232, opened 2020-03-13
feature/[STREAMP-1951]-include-schema-key-in-streams

dccarlos:
# stream-registry PR

Include schema key in stream notification events

### Changed
* A new field was added for `AvroStream record` in stream-registry-notification schema
* Unit tests were modified to verify and test the new field

# PR Checklist Forms

- [X] CHANGELOG.md updated
- [ ] Reviewer assigned
- [X] PR assigned (presumably to submitter)


notification-support/src/main/java/com/expediagroup/streamplatform/streamregistry/core/events/NotificationEventUtils.java
```diff
@@ -51,6 +52,18 @@ public static AvroKey toAvroKeyRecord(Schema schema) {
         .build();
   }
 
+  public static AvroKey toAvroKeyRecord(SchemaKey schemaKey) {
+    return Optional.ofNullable(schemaKey)
+        .filter(sk -> sk.getName() != null)
+        .map(sk -> AvroKey.newBuilder()
+            .setId(sk.getName())
+            .setVersion(null)
```
rshotels: Do we really need to set null? I think I would prefer we don't if null is the default.

notification-support/src/main/java/com/expediagroup/streamplatform/streamregistry/core/events/NotificationEventUtils.java
```diff
@@ -51,6 +52,18 @@ public static AvroKey toAvroKeyRecord(Schema schema) {
         .build();
   }
 
+  public static AvroKey toAvroKeyRecord(SchemaKey schemaKey) {
```
rshotels: is it _ever_ valid for the avro key to be null? Or should we throw an exception / not persist.
I don't know the answer to that.

nahguam: No, it should never be null. Can we avoid the `Optional` here?

notification-support/src/main/java/com/expediagroup/streamplatform/streamregistry/core/events/NotificationEventUtils.java
nahguam: Good spot. I completely missed this.

The version field shouldn't be present. It's not what I intended with the general purpose key for all types.

```avdl
  record AvroKey {
    string id;
    union {null, int} version = null; // This field shouldn't be present
    AvroKeyType type;
    union {null, AvroKey} parent;
  }
```

I also see below that the parent is not being set. In this case the parent is the domain. 

Examples:

```yaml
schema:
  id: "schema_1"
  type: "SCHEMA"
  parent:
    id: "domain_1"
    type: "DOMAIN"
    parent: null
```

```yaml
stream:
  id: "1"
  type: "STREAM_VERSION"
  parent:
    id: "stream_1"
    type: "STREAM"
    parent:
      id: "domain_1"
      type: "DOMAIN"
      parent: null
```

This is an existing bug that we can probably address separately. Thoughts?

notification-support/src/main/java/com/expediagroup/streamplatform/streamregistry/core/events/NotificationEventUtils.java
dccarlos: Sure, since I added new field I didn't want to break the schema compatibility. I assumed it is currently `BACKWARD` compatible.

That's why I used `Optional.ofNullable`

notification-support/src/main/java/com/expediagroup/streamplatform/streamregistry/core/events/NotificationEventUtils.java
dccarlos: Got it, now I got the purpose of `"STREAM_VERSION"`. I'll remove the version field and use the example structure.

dccarlos: I ran some samples in my local and these were my results. I think I got the concept :)
```
// Sample key for STREAM
{
  "id": "1",
  "type": "STREAM_VERSION",
  "parent": {
    "com.expediagroup.streamplatform.streamregistry.avro.AvroKey": {
      "id": "demonix_05",
      "type": "STREAM",
      "parent": {
        "com.expediagroup.streamplatform.streamregistry.avro.AvroKey": {
          "id": "dh_test_01",
          "type": "DOMAIN",
          "parent": null
        }
      }
    }
  }
}

// Sample key for SCHEMA
{
  "id": "ccordero_schema",
  "type": "SCHEMA",
  "parent": {
    "com.expediagroup.streamplatform.streamregistry.avro.AvroKey": {
      "id": "my_test_domain",
      "type": "DOMAIN",
      "parent": null
    }
  }
}
```

notification-support/src/main/java/com/expediagroup/streamplatform/streamregistry/core/events/NotificationEventUtils.java
dccarlos: Removed

notification-support/src/main/java/com/expediagroup/streamplatform/streamregistry/core/events/NotificationEventUtils.java
```diff
@@ -91,13 +114,25 @@ public static AvroKey toAvroKeyRecord(Stream stream) {
...
     val version = stream.getKey().getVersion();
+    val domainName = stream.getKey().getDomain();
 
-    return AvroKey.newBuilder()
-        .setId(name)
-        .setVersion(version)
+    var domainKey = AvroKey.newBuilder()
+        .setId(domainName)
+        .setType(AvroKeyType.DOMAIN)
         .setParent(null)
```
rshotels: Again do we need to set null?

notification-support/src/test/java/com/expediagroup/streamplatform/streamregistry/core/events/CustomSchemaMethodsLoadingTest.java
```diff
@@ -100,10 +100,18 @@ public void having_defined_custom_parser_methods_verify_they_execute_properly()
   }
 
   public static AvroKey myCustomKey(Schema schema) {
-    val key = AvroKey.newBuilder()
-        .setId(schema.getKey().getName())
-        .setVersion(null)
+    val name = schema.getKey().getName();
```
rshotels:
```suggestion
    var key = AvroKey.newBuilder()
        .setId(schema.getKey().getName())
        .setParent(AvroKey.newBuilder()
              .setId(schema.getKey().getDomain())
              .setType(AvroKeyType.DOMAIN)
              .build())
        .setType(AvroKeyType.SCHEMA)
        .build();
```
I would consider nesting as above, but I guess it is personal preference.

notification-support/src/main/java/com/expediagroup/streamplatform/streamregistry/core/events/NotificationEventUtils.java
dccarlos: Yes, because avro schema doesn't have a default value. It will cause an error if it is not set explicitly.
```
    union {null, AvroKey} parent;
```

dccarlos: A default value could set explicitly though. But I'm preserving the original schema design.
```
union {null, AvroKey} parent = null; 
```

rshotels: In that case I don't really see how a builder is the appropriate construction mechanism.

can we just have an all args constructor and be done with it?

dccarlos: @nahguam do you see any trouble setting a default null for the parent?:
```
union {null, AvroKey} parent = null; 
```

notification-support/src/main/java/com/expediagroup/streamplatform/streamregistry/core/events/NotificationEventUtils.java
```diff
@@ -35,18 +35,39 @@
...
 import com.expediagroup.streamplatform.streamregistry.model.Tag;
+import com.expediagroup.streamplatform.streamregistry.model.keys.SchemaKey;
 
 @Slf4j
 public class NotificationEventUtils {
   public static AvroKey toAvroKeyRecord(Schema schema) {
     validateSchemaKey(schema);
 
     val name = schema.getKey().getName();
+    val domainName = schema.getKey().getDomain();
```
cricket007:
```suggestion
    val key = schema.getKey();
    val domainName = key.getDomain();
```

notification-support/src/main/java/com/expediagroup/streamplatform/streamregistry/core/events/NotificationEventUtils.java
```diff
@@ -91,13 +112,24 @@ public static AvroKey toAvroKeyRecord(Stream stream) {
 
     val name = stream.getKey().getName();
     val version = stream.getKey().getVersion();
+    val domainName = stream.getKey().getDomain();
```
cricket007:
```suggestion
    val key = stream.getKey();
    val domainName = key.getDomain();
```

notification-support/src/main/java/com/expediagroup/streamplatform/streamregistry/core/events/NotificationEventUtils.java
```diff
@@ -151,6 +185,12 @@ private static void validateSchemaKey(Schema schema) {
     checkNotNull(schema.getKey().getDomain(), canNotBeNull("key's domain"));
   }
 
+  private static void validateSchemaKey(SchemaKey schemaKey) {
+    checkNotNull(schemaKey, canNotBeNull("schema key"));
```
cricket007: FWIW, this can be `Objects.requireNonNull` if you want to remove Guava dep

dccarlos: That's a good idea, even if guava is transitively available in the module.

notification-support/src/main/java/com/expediagroup/streamplatform/streamregistry/core/events/NotificationEventUtils.java
dccarlos: You're right, I also modified others ones

notification-support/pom.xml
```diff
@@ -24,10 +24,6 @@
       <groupId>org.apache.avro</groupId>
       <artifactId>avro</artifactId>
     </dependency>
-    <dependency>
-      <groupId>com.google.guava</groupId>
-      <artifactId>guava</artifactId>
-    </dependency>
```
cricket007: 🔥

PR #237, opened 2020-03-19
performance improvement using jpa example

nahguam:
# stream-registry PR

performance improvement using jpa example

# PR Checklist Forms

- [ ] CHANGELOG.md updated
- [ ] Reviewer assigned

### DIFF
--- a/core/src/main/java/com/expediagroup/streamplatform/streamregistry/core/repositories/CachingJpaRepository.java
+++ b/core/src/main/java/com/expediagroup/streamplatform/streamregistry/core/repositories/CachingJpaRepository.java
@@ -15,26 +15,31 @@
  */
 package com.expediagroup.streamplatform.streamregistry.core.repositories;
 
-import java.util.Optional;
+import static org.hibernate.annotations.QueryHints.CACHEABLE;
 
+import java.util.List;
+import java.util.Optional;
 import javax.persistence.QueryHint;
 
+import org.springframework.data.domain.Example;
+import org.springframework.data.jpa.repository.JpaRepository;
 import org.springframework.data.jpa.repository.QueryHints;
-import org.springframework.data.repository.CrudRepository;
 import org.springframework.data.repository.NoRepositoryBean;
 
 @NoRepositoryBean
-public interface CachedCrudRepository<T, ID> extends CrudRepository<T, ID> {
+public interface CachingJpaRepository<T, ID> extends JpaRepository<T, ID> {
+  @QueryHints(@QueryHint(name = CACHEABLE, value = "true"))
+  Optional<T> findById(ID id);
 
-  @QueryHints(@QueryHint(name = org.hibernate.annotations.QueryHints.CACHEABLE, value = "true"))
-  Optional<T> findById(ID var1);
+  @QueryHints(@QueryHint(name = CACHEABLE, value = "true"))
+  boolean existsById(ID id);
 
-  @QueryHints(@QueryHint(name = org.hibernate.annotations.QueryHints.CACHEABLE, value = "true"))
-  boolean existsById(ID var1);
+  @QueryHints(@QueryHint(name = CACHEABLE, value = "true"))
+  List<T> findAll();
 
-  @QueryHints(@QueryHint(name = org.hibernate.annotations.QueryHints.CACHEABLE, value = "true"))
-  Iterable<T> findAll();
+  @QueryHints(@QueryHint(name = CACHEABLE, value = "true"))
+  List<T> findAllById(Iterable<ID> ids);
 
-  @QueryHints(@QueryHint(name = org.hibernate.annotations.QueryHints.CACHEABLE, value = "true"))
-  Iterable<T> findAllById(Iterable<ID> var1);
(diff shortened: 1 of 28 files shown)


core/src/main/java/com/expediagroup/streamplatform/streamregistry/core/services/ProducerBindingService.java
```diff
@@ -79,14 +80,25 @@ public void delete(ProducerBinding producerBinding) {
...
+    ProducerBinding example = new ProducerBinding(
+        new ProducerBindingKey(
+            key.getStreamDomain(),
+            key.getStreamName(),
+            key.getStreamVersion(),
+            key.getZone(),
+            null,
+            key.getName()
+        ), null, null);
+    return producerBindingRepository.findAll(Example.of(example)).stream().findFirst();
```
nahguam: This is where the magic happens.

rshotels: So this produces an actual HQL/ SQL query which matches the example, right?

core/src/main/java/com/expediagroup/streamplatform/streamregistry/core/services/ProducerBindingService.java
```diff
@@ -79,14 +80,25 @@ public void delete(ProducerBinding producerBinding) {
   }
 
   public Iterable<ProducerBinding> findAll(Predicate<ProducerBinding> filter) {
-    return stream(producerBindingRepository.findAll().spliterator(), false)
-        .filter(r -> filter.test(r))
-        .collect(Collectors.toList());
+    return producerBindingRepository.findAll().stream().filter(filter).collect(toList());
```
rshotels: Might be possible to use the 'Example' process if the filter has any non-regex fields, then apply the Predicate for the regex. Something for tomorrow?

model/src/main/java/com/expediagroup/streamplatform/streamregistry/model/ConsumerBinding.java
```diff
@@ -36,7 +40,5 @@
   @EmbeddedId
   private ConsumerBindingKey key;
   private Specification specification;
-  private String consumerId; //Resolves to Consumer
-  private String bindingId; //Resolves to StreamBinding
```
rshotels: Are you sure these can come out?

core/src/main/java/com/expediagroup/streamplatform/streamregistry/core/services/ProducerBindingService.java
nahguam: I imagine so.

core/src/main/java/com/expediagroup/streamplatform/streamregistry/core/repositories/CachingJpaRepository.java
```diff
@@ -15,26 +15,31 @@
...
 
-  @QueryHints(@QueryHint(name = org.hibernate.annotations.QueryHints.CACHEABLE, value = "true"))
-  boolean existsById(ID var1);
+  @QueryHints(@QueryHint(name = CACHEABLE, value = "true"))
+  List<T> findAll();
 
-  @QueryHints(@QueryHint(name = org.hibernate.annotations.QueryHints.CACHEABLE, value = "true"))
-  Iterable<T> findAll();
+  @QueryHints(@QueryHint(name = CACHEABLE, value = "true"))
+  List<T> findAllById(Iterable<ID> ids);
```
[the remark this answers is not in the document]
nahguam: I just changed the param name to the same as it is in the super interface.

core/src/main/java/com/expediagroup/streamplatform/streamregistry/core/services/ProducerBindingService.java
nahguam: Everything is a regex string apart from the stream version which is an integer. May be worth exploring but it's not low hanging fruit and also separately from this PR.

model/src/main/java/com/expediagroup/streamplatform/streamregistry/model/ConsumerBinding.java
nahguam: They're not there on the `ProducerBindingKey` and that works, and the build tests pass 👍

rshotels: ok

core/src/main/java/com/expediagroup/streamplatform/streamregistry/core/services/ProducerBindingService.java
rshotels: Just putting it out there before I forget.

core/src/main/java/com/expediagroup/streamplatform/streamregistry/core/repositories/CachingJpaRepository.java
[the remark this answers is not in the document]
rshotels: Well ok then

PR #241, opened 2020-04-07
Feature/streamp 1431 split model into two

rshotels:
# stream-registry PR 1431 split model into two

Split internal representation into Model and Data packages


core/src/main/java/com/expediagroup/streamplatform/streamregistry/core/handlers/HandlerService.java
```diff
@@ -48,7 +48,7 @@
 @Component
 @Slf4j
 public class HandlerService {
-  private final Map<Key, Handler<?>> handlers;
```
nahguam, line 51: Please leave the generic wildcard

core/src/main/java/com/expediagroup/streamplatform/streamregistry/core/handlers/HandlerService.java
```diff
@@ -79,7 +79,7 @@ private static void check(Set<Class> handlerTargets, Class target) {
     checkArgument(handlerTargets.contains(target), "No Handlers for " + target.getSimpleName() + " defined");
   }
 
-  private <T extends ManagedType> Handler<T> getHandler(T managedType) {
+  private <T extends Modeled> Handler<T> getHandler(T managedType) {
```
nahguam, line 82: Not keen on `Modeled`. I assume the 'data' equivalent is still `ManagedType`?
How about `Entity` for the (public facing) service layer (nice, simple, clear) and `DataEntity` for the repository layer?

core/src/main/java/com/expediagroup/streamplatform/streamregistry/core/services/ConsumerBindingService.java
```diff
@@ -43,35 +46,47 @@
...
+    consumerBindingData.setSpecification(convertToData(handlerService.handleInsert(consumerBindingModel)));
+    ConsumerBinding out = convertToModel(consumerBindingRepository.save(consumerBindingData));
+    consumerBindingServiceEventEmitter.emitEventOnProcessedEntity(EventType.CREATE, out);
+    return Optional.ofNullable(out);
   }
 
   public Optional<ConsumerBinding> read(ConsumerBindingKey key) {
-    return consumerBindingRepository.findById(key);
+    var data = consumerBindingRepository.findById(convertToData(key));
+    return data.isPresent() ? Optional.of(convertToModel(data.get())) : Optional.empty();
```
nahguam, line 63:
```suggestion
    return data.map(DataToModel::convertToModel);
```

core/src/main/java/com/expediagroup/streamplatform/streamregistry/core/services/ConsumerBindingService.java
```diff
@@ -43,35 +46,47 @@
...
   }
 
   public Iterable<ConsumerBinding> readAll() {
-    return consumerBindingRepository.findAll();
+    ArrayList out = new ArrayList();
```
nahguam, line 67: `Iterable` is a pain in the java8+ world...

FYI, since the switch from `CrudRepository` to `JpaRepository`, `findAll` now returns `List` so you could go streamy here (I assumed being `Iterable` before is the reason you've gone with the classic for loop

core/src/main/java/com/expediagroup/streamplatform/streamregistry/core/services/ConsumerBindingService.java
```diff
@@ -15,11 +15,13 @@
  */
 package com.expediagroup.streamplatform.streamregistry.core.services;
 
+import static com.expediagroup.streamplatform.streamregistry.DataToModel.convertToModel;
```
nahguam, line 18: proper injected collaborators, rather than static utility methods, would simplify testing. Here you'd be able to pass mocks around to verify composition and leave the conversion testing to the converter class.

core/src/main/java/com/expediagroup/streamplatform/streamregistry/core/services/InfrastructureService.java
```diff
@@ -41,34 +45,46 @@
   private final NotificationEventEmitter<Infrastructure> infrastructureServiceEventEmitter;
 
   public Optional<Infrastructure> create(Infrastructure infrastructure) throws ValidationException {
-    if (infrastructureRepository.findById(infrastructure.getKey()).isPresent()) {
+    var data = convertToData(infrastructure);
```
nahguam, line 48: All the services do look very similar with all the conversion logic (as you've already mentioned), but we knew that already before. Aside from `Stream` entities (that have `Schemas`) all others are likely to be basically identical. I'm pretty sure we could have one default service impl, one for `Stream` using the default as parent with override for to handle the schema.

core/src/test/java/com/expediagroup/streamplatform/streamregistry/core/events/NotificationEventEmitterConsumerBindingServiceTest.java
```diff
@@ -1,222 +0,0 @@
-/**
```
nahguam, line 1: why have all these tests been deleted?

notification-support/src/test/java/com/expediagroup/streamplatform/streamregistry/core/events/NotificationEventListenerKafkaIntegrationTest.java
```diff
@@ -243,39 +249,31 @@ public static String getSubjectName(String topic, boolean isKey) {
...
-    return NotificationEvent.<Schema>builder()
+    return NotificationEvent.<Schema> builder()
         .entity(schema)
         .source(source)
         .eventType(eventType)
         .build();
   }
 
   public static Schema getDummySchema() {
-    val name = Instant.now().toString();
```
nahguam, line 254: Why remove `lombok.val`? Strictly I guess we should use `var` being java 11

stream-registry-data/src/main/java/com/expediagroup/streamplatform/streamregistry/ModelToData.java
```diff
@@ -0,0 +1,146 @@
...
+import com.expediagroup.streamplatform.streamregistry.data.keys.ProducerKey;
+import com.expediagroup.streamplatform.streamregistry.data.keys.SchemaKey;
+import com.expediagroup.streamplatform.streamregistry.data.keys.StreamBindingKey;
+import com.expediagroup.streamplatform.streamregistry.data.keys.StreamKey;
+import com.expediagroup.streamplatform.streamregistry.data.keys.ZoneKey;
+
+public class ModelToData {
+
+  public static ConsumerKey convertToData(com.expediagroup.streamplatform.streamregistry.model.keys.ConsumerKey in) {
+    return new ConsumerKey(in.getStreamDomain(), in.getStreamName(), in.getStreamVersion(), in.getZone(), in.getName());
```
nahguam, line 50: [bull](https://github.com/HotelsDotCom/bull) could do most of this automatically, you'd only need to provide handlers for the config/status serde.

It's already being dependency managed as was being used previously https://github.com/ExpediaGroup/stream-registry/blob/410a45bc01cbacf1c93aab60324f8f066efa92e8/pom.xml#L135

core/src/main/java/com/expediagroup/streamplatform/streamregistry/core/handlers/HandlerService.java
rshotels: done

core/src/main/java/com/expediagroup/streamplatform/streamregistry/core/handlers/HandlerService.java
rshotels: We can't change ManagedType, and I don't feel that Entity really conveys anything. Everything is an Entity.

Can we leave this for now?

core/src/main/java/com/expediagroup/streamplatform/streamregistry/core/services/ConsumerBindingService.java
rshotels: Agreed, but returning iterable is. the existing functionality.

core/src/main/java/com/expediagroup/streamplatform/streamregistry/core/services/ConsumerBindingService.java
rshotels: Thats true but unit tests against DataToModel were written with no difficulty. If it is used in other classes, unit tests can still be written and the DataToModel functions have no dependencies.

Can you give an example if what would be solved. I agree in principal but don't see where it helps us in this case.

core/src/main/java/com/expediagroup/streamplatform/streamregistry/core/services/InfrastructureService.java
rshotels: I believe that is true but would prefer to dot it as a separate ticket. Have actually tried this and it involves considerable code and test changes.

core/src/test/java/com/expediagroup/streamplatform/streamregistry/core/events/NotificationEventEmitterConsumerBindingServiceTest.java
rshotels: Because they massively overreach what they need to test, and don't really prove anything.

notification-support/src/test/java/com/expediagroup/streamplatform/streamregistry/core/events/NotificationEventListenerKafkaIntegrationTest.java
rshotels: because it adds nothing. All the values are Strings and all the others are identified as such.

stream-registry-data/src/main/java/com/expediagroup/streamplatform/streamregistry/ModelToData.java
rshotels: Done.

core/src/main/java/com/expediagroup/streamplatform/streamregistry/core/handlers/HandlerService.java
nahguam: Once this interface is extracted and being used externally we'll never come back and change it so I'd prefer we do it now.

Why can't we change `ManagedType`? `Entity` conveys more than `ManagedType` which is a bit cryptic. Says nothing about. At least Entity says what it is. And the model and data interfaces should at least have some equivalent in their naming - `ManagedType` and `Modeled` have none.

core/src/main/java/com/expediagroup/streamplatform/streamregistry/core/services/ConsumerBindingService.java
nahguam: it means you have to test the conversion twice - once in the converter itself and again where it's used. But we have no isolated units tests for `ConsumerBindingService` so I can't give you an example.

core/src/main/java/com/expediagroup/streamplatform/streamregistry/core/services/ConsumerBindingService.java
nahguam: `Iterable` was only chosen before because the Repository exposed it. We now have `List` exposed. Everywhere else is very streamy in style. This is the only place it is not.

core/src/main/java/com/expediagroup/streamplatform/streamregistry/core/services/InfrastructureService.java
nahguam: Yes can address this separately.

core/src/main/java/com/expediagroup/streamplatform/streamregistry/core/handlers/HandlerService.java
```diff
@@ -94,15 +94,15 @@ private static void check(Set<Class> handlerTargets, Class target) {
...
+  public <T extends Modeled> Specification handleInsert(T managedType) {
     try {
       return getHandler(managedType).handleInsert(managedType);
     } catch (Exception e) {
       throw new ValidationException(e);
     }
   }
 
-  public Specification handleUpdate(ManagedType managedType, ManagedType existing) {
+  public Specification handleUpdate(Modeled managedType, Modeled existing) {
```
nahguam, line 105: Parameter names don't match the type. I believe `Entity` or similar variant would convey more information about what it is here. `Modeled` (and `ManagedType`) are too abstract.

core/src/main/java/com/expediagroup/streamplatform/streamregistry/core/handlers/HandlerService.java
rshotels: ManagedType is an interface in the persistence package, we don't get to change that.

core/src/main/java/com/expediagroup/streamplatform/streamregistry/core/services/InfrastructureService.java
```diff
@@ -41,34 +45,46 @@
...
+    data.setSpecification(convertToData(handlerService.handleInsert(infrastructure)));
+    Infrastructure out = convertToModel(infrastructureRepository.save(data));
+    infrastructureServiceEventEmitter.emitEventOnProcessedEntity(EventType.CREATE, out);
+    return Optional.ofNullable(out);
   }
 
   public Optional<Infrastructure> read(InfrastructureKey key) {
-    return infrastructureRepository.findById(key);
+    var data = infrastructureRepository.findById(convertToData(key));
+    return data.isPresent() ? Optional.of(convertToModel(data.get())) : Optional.empty();
```
nahguam, line 61: `Optional.map`

core/src/main/java/com/expediagroup/streamplatform/streamregistry/core/services/ProducerBindingService.java
```diff
@@ -43,34 +48,45 @@
...
+    data.setSpecification(convertToData(handlerService.handleInsert(producerBinding)));
+    ProducerBinding out = DataToModel.convertToModel(producerBindingRepository.save(data));
+    producerBindingServiceEventEmitter.emitEventOnProcessedEntity(EventType.CREATE, out);
+    return Optional.ofNullable(out);
   }
 
   public Optional<ProducerBinding> read(ProducerBindingKey key) {
-    return producerBindingRepository.findById(key);
+    var data = producerBindingRepository.findById(convertToData(key));
+    return data.isPresent() ? Optional.of(DataToModel.convertToModel(data.get())) : Optional.empty();
```
nahguam, line 64: `Optional.map`

notification-support/src/test/java/com/expediagroup/streamplatform/streamregistry/core/events/NotificationEventListenerKafkaIntegrationTest.java
nahguam: But you have yourself used Java's `var` elsewhere.

core/src/test/java/com/expediagroup/streamplatform/streamregistry/core/events/NotificationEventEmitterConsumerBindingServiceTest.java
nahguam: I think completely deleting them is going too far.

notification-support/src/test/java/com/expediagroup/streamplatform/streamregistry/core/events/NotificationEventListenerKafkaIntegrationTest.java
rshotels: Ok... but in this context we have a several Strings and a var for no reason.

core/src/main/java/com/expediagroup/streamplatform/streamregistry/core/handlers/HandlerService.java
rshotels: Happy to hear suggestions of what would convey the intent better than Modeled.

To me a 'Modeled' Object is one that exists in the core Model. It has a state and a specification. What should we call it instead?

core/src/test/java/com/expediagroup/streamplatform/streamregistry/core/events/NotificationEventEmitterConsumerBindingServiceTest.java
rshotels: They also don't work. But thats because they are overreaching, not because the functionality under test is broken.

core/src/main/java/com/expediagroup/streamplatform/streamregistry/core/services/ConsumerBindingService.java
rshotels: Changed to a List in all services.
Also removed unused 'findAll' methods

core/src/main/java/com/expediagroup/streamplatform/streamregistry/core/handlers/HandlerService.java
rshotels: will rename the parameters. Good spot.

core/src/main/java/com/expediagroup/streamplatform/streamregistry/core/handlers/HandlerService.java
rshotels: Happy to hear suggestions of what would convey the intent better than Modeled.
To me a 'Modeled' Object is one that exists in the core Model. It has a state and a specification. What should we call it instead?

core/src/main/java/com/expediagroup/streamplatform/streamregistry/core/services/ConsumerBindingService.java
rshotels: ModelToData and DataToModel are both now injected Components

notification-support/src/test/java/com/expediagroup/streamplatform/streamregistry/core/events/NotificationEventListenerKafkaIntegrationTest.java
rshotels: I'm not arguing that val/var add nothing. I'm saying this specific use of val  adds nothing, if anything it is more confusing since all the other variables are defined as Strings except this one (which is also a String)

core/src/main/java/com/expediagroup/streamplatform/streamregistry/core/services/InfrastructureService.java
rshotels: That is much nicer. Done.

core/src/main/java/com/expediagroup/streamplatform/streamregistry/core/handlers/HandlerService.java
rshotels: Changed to Entity and DataEntity

core/src/test/java/com/expediagroup/streamplatform/streamregistry/core/events/NotificationEventEmitterConsumerBindingServiceTest.java
rshotels: I'm looking into restoring these tests, but they look more suited to an ITest scenario to me.

data/src/main/java/com/expediagroup/streamplatform/streamregistry/data/Specified.java
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.expediagroup.streamplatform.streamregistry.data;
+
+@Deprecated // model only
```
nahguam, line 18: Can this be removed then?

data/src/main/java/com/expediagroup/streamplatform/streamregistry/data/Stated.java
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.expediagroup.streamplatform.streamregistry.data;
+
+@Deprecated // model only
```
nahguam, line 18: Can this be removed then?

data/src/main/java/com/expediagroup/streamplatform/streamregistry/data/Status.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+package com.expediagroup.streamplatform.streamregistry.data;
+
+import javax.persistence.Column;
+import javax.persistence.Embeddable;
+import javax.persistence.Lob;
+
+import lombok.Data;
+
+@Deprecated // do we need this since it holds one field?
```
nahguam, line 24: We're going to add support for multiple agents to save status, basically a map of status, keyed by name.

Please can you remove the @Deprecated annotation

data/src/main/java/com/expediagroup/streamplatform/streamregistry/data/Stated.java
rshotels: done

data/src/main/java/com/expediagroup/streamplatform/streamregistry/data/Status.java
rshotels: done

data/src/main/java/com/expediagroup/streamplatform/streamregistry/data/Specified.java
rshotels: Done

core/src/test/java/com/expediagroup/streamplatform/streamregistry/core/events/NotificationEventEmitterConsumerBindingServiceTest.java
rshotels: Tests restored

PR #246, opened 2020-05-15
state library

nahguam:
# stream-registry PR

# PR Checklist Forms

- [ ] CHANGELOG.md updated
- [ ] Reviewer assigned


state/avro/src/main/java/com/expediagroup/streamplatform/streamregistry/state/avro/AvroObjectDeserializer.java
```diff
@@ -0,0 +1,97 @@
...
+import com.fasterxml.jackson.core.JsonToken;
+import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.JsonDeserializer;
+
+@Slf4j
+class AvroObjectDeserializer extends JsonDeserializer<AvroObject> {
+  @Override
+  public AvroObject deserialize(JsonParser p, DeserializationContext ctxt) throws IOException {
+    if (p.currentToken() != START_OBJECT) {
+      throw new IllegalStateException("Unexpected start token: " + p.currentToken());
```
rshotels, line 39: Won't deserialising do similar on failure?

state/graphql-sender/pom.xml
```diff
@@ -0,0 +1,116 @@
...
+              <outputDirectory>${project.build.directory}/generated-sources/graphql-client</outputDirectory>
+              <introspectionFile>${introspectionFile}</introspectionFile>
+              <rootPackageName>${clientPackage}</rootPackageName>
+              <customTypeMap>
+                <ObjectNode>com.fasterxml.jackson.databind.node.ObjectNode</ObjectNode>
+              </customTypeMap>
+              <suppressRawTypesWarning>true</suppressRawTypesWarning>
+            </configuration>
+          </execution>
+          <!--          <execution>-->
```
rshotels, line 93: Can we remove this if it isn't in use?

state/kafka-sender/src/test/java/com/expediagroup/streamplatform/streamregistry/state/kafka/KafkaEventReceiverTest.java
```diff
@@ -0,0 +1,95 @@
...
+
+import lombok.extern.slf4j.Slf4j;
+
+
+
+
+@Slf4j
+public class KafkaEventReceiverTest {
+
+//  @Rule
```
rshotels, line 28: Can we either enable or remove this test.

state/graphql-sender/pom.xml
nahguam: Yes, I left it there in case I do a full IT.

state/avro/src/main/java/com/expediagroup/streamplatform/streamregistry/state/avro/AvroObjectDeserializer.java
nahguam: Not sure I follow the question

state/kafka-sender/src/test/java/com/expediagroup/streamplatform/streamregistry/state/kafka/KafkaEventReceiverTest.java
nahguam: Sorry, accidentally staged. Will get cleaned up when I pushed the tests.

nahguam: done

state/graphql-sender/pom.xml
nahguam: done

state/avro/src/main/java/com/expediagroup/streamplatform/streamregistry/state/avro/AvroObjectDeserializer.java
rshotels: If the first token isn't START_OBJECT I would expect deserialisation to fail anyway.
Not that important, just curious.

nahguam: Well you could do something silly like:

`mapper.convertValue(primitive, AvroObject.class)`

or

`mapper.convertValue(array, AvroObject.class)`

state/kafka-receiver/src/main/java/com/expediagroup/streamplatform/streamregistry/state/kafka/KafkaEventReceiver.java
```diff
@@ -0,0 +1,170 @@
...
+
+import com.expediagroup.streamplatform.streamregistry.state.EventCorrelator;
+import com.expediagroup.streamplatform.streamregistry.state.EventReceiver;
+import com.expediagroup.streamplatform.streamregistry.state.EventReceiverListener;
+import com.expediagroup.streamplatform.streamregistry.state.avro.AvroConverter;
+import com.expediagroup.streamplatform.streamregistry.state.avro.AvroKey;
+import com.expediagroup.streamplatform.streamregistry.state.avro.AvroValue;
+
+@Slf4j
+@RequiredArgsConstructor(access = PACKAGE)
```
dccarlos, line 59: Did you consider to use `Kafka Streams`? It has support for views (`KTables`, `GlobalKTables`) and you can expose the stores through an API, correlators (`Joiners`), you wouldn't deal directly with producers and consumers and there is a clear way to deal with more than one partition...

The stream-registry to datahub mappers works in that way. The last version is going to have three stores, this is the design document:

https://pages.github.expedia.biz/eg-data-platform/metadata-ingestion-streams/datahub/design/kafka-streams-modeling/

state/kafka-receiver/src/main/java/com/expediagroup/streamplatform/streamregistry/state/kafka/KafkaEventReceiver.java
```diff
@@ -0,0 +1,170 @@
...
+    consumer.seekToBeginning(topicPartitions);
+
+    var loaded = false;
+    if (endOffset == 0L) {
+      progressLogger.cancel(true);
+      log.info("Loading complete. Empty topic.");
+      listener.onEvent(LOAD_COMPLETE);
+      loaded = true;
+    }
+
```
dccarlos, line 114: Maybe it is a good idea to define some lifecycle for this long running task for monitoring purposes and a more robust cancelation policy by using—for example—`RunnableFuture`

state/kafka-receiver/src/main/java/com/expediagroup/streamplatform/streamregistry/state/kafka/KafkaEventReceiver.java
nahguam: Sorry, I removed to internal link you provided. 

Yes I did. It works for simple use cases, but gets convoluted if your use case doesn't exactly fit the mould. For example we need a `GlobalKTable` so we get the whole dataset in each instance and so it waits for the table to be completely loaded before allowing querying of the KV store. But the topic has to be in the exact KV structure that the KV store needs or you have to build out a transform step into a second topic that the table can source from. I want to keep it simple - 1 topic and avoid coupling to Kstreams impl detail patterns.

state/kafka-receiver/src/main/java/com/expediagroup/streamplatform/streamregistry/state/kafka/KafkaEventReceiver.java
nahguam: I originally started out with a reactive streams approach.

The current approach works simply and nicely in a spring boot app as it's closeable so spring already knows how to deal with it within the application context.

state/api/src/main/java/com/expediagroup/streamplatform/streamregistry/state/internal/EventCorrelator.java
```diff
@@ -0,0 +1,55 @@
...
+import com.expediagroup.streamplatform.streamregistry.state.EventSender;
+import com.expediagroup.streamplatform.streamregistry.state.model.event.Event;
+
+/**
+ * Provides a mechanism for an {@link EventSender} to coordinate with an {@link EventReceiver}. This allows
+ * it to be able to wait until the receiver confirms receipt before completing the {@link EventSender#send(Event)}
+ * call. Intended for use internally within Stream Registry itself only.
+ */
+public interface EventCorrelator {
+  String CORRELATION_ID = "correlationId";
```
rshotels, line 30: What is the purpose of this value being defined here?

state/api/src/main/java/com/expediagroup/streamplatform/streamregistry/state/model/specification/Tag.java
```diff
@@ -0,0 +1,25 @@
...
+ */
+package com.expediagroup.streamplatform.streamregistry.state.model.specification;
+
+import lombok.NonNull;
+import lombok.Value;
+
+@Value
+public class Tag {
+  @NonNull String name;
+  @NonNull String value;
```
rshotels, line 24: I think potentially a Tag value should be nullable, for example an Entity could be tagged as being 'UNUSED'.

state/avro/src/main/java/com/expediagroup/streamplatform/streamregistry/state/avro/AvroObjectSerializer.java
```diff
@@ -0,0 +1,76 @@
...
+    } else if (value instanceof Long) {
+      g.writeNumber((long) value);
+    } else if (value instanceof Integer) {
+      g.writeNumber((int) value);
+    } else if (value instanceof Boolean) {
+      g.writeBoolean((boolean) value);
+    } else if (value == null) {
+      g.writeNull();
+    }
+  }
```
rshotels, line 58: Is it possible to get here having written no output? Perhaps we should trap that?

state/api/src/main/java/com/expediagroup/streamplatform/streamregistry/state/internal/EventCorrelator.java
nahguam: Both the kafka receiver and sender need it and I'm loath to create a constants class just for that.

state/api/src/main/java/com/expediagroup/streamplatform/streamregistry/state/model/specification/Tag.java
nahguam: I think you're right - it's nullable in the graphql api. I'll have another look

state/avro/src/main/java/com/expediagroup/streamplatform/streamregistry/state/avro/AvroObjectSerializer.java
nahguam: good spot

PR #258, opened 2020-07-29
Feature/state api authentication support

a-nersharma:
# stream-registry PR

Added Authentication support for state api

### Added
* Authentication support for state api

# PR Checklist Forms

- [x] CHANGELOG.md updated
- [x] Reviewer assigned


state/graphql-sender/src/main/java/com/expediagroup/streamplatform/streamregistry/state/graphql/DefaultApolloClientFactory.java
```diff
@@ -29,24 +29,50 @@
...
+
   @Override
   public ApolloClient create() {
-    var builder = builder()
-        .okHttpClient(new OkHttpClient.Builder().build());
+    OkHttpClient okHttpClient;
+
+    if (credentials != null) {
+      okHttpClient = new OkHttpClient.Builder()
+          .authenticator((route, response) -> response.request().newBuilder()
```
nahguam, line 54: We don't need the authenticator AND the interceptor - just the interceptor. The code example in the ticket was just illustrating both ways of doing it.

state/graphql-sender/src/main/java/com/expediagroup/streamplatform/streamregistry/state/graphql/Credentials.java
```diff
@@ -0,0 +1,32 @@
...
+
+import lombok.RequiredArgsConstructor;
+
+@RequiredArgsConstructor
+public class Credentials {
+  private final String userName;
+  private final String password;
+
+  public String basic() {
+    return Base64.getEncoder().encodeToString((userName + ":" + password).getBytes(UTF_8));
```
nahguam, line 30: Apollo has a `Credentials` class that already does the encoding. Prefer not to have to maintain another class that does the same thing. Perhaps just have 2 parameters in `DefaultApolloClientFactory`?

state/graphql-sender/src/main/java/com/expediagroup/streamplatform/streamregistry/state/graphql/DefaultApolloClientFactory.java
```diff
@@ -29,24 +29,50 @@
...
+          .addInterceptor(chain -> chain.proceed(chain.request().newBuilder()
+              .header("Authorization", credentials.basic())
+              .build()))
+          .build();
+    } else {
+      okHttpClient = new OkHttpClient.Builder().build();
+    }
+
+    var builder = builder().okHttpClient(okHttpClient);
     configurer.accept(builder);
```
nahguam, line 66: Because the `OkHttpClient` logic comes before the `configurer` it's possible the user could override it. Perhaps let's get rid of the configurer for now. We can deal with being able to configure other aspects of the `OkHttpClient` later if we need to.

Maybe something like:

```
    var okHttpClientBuilder = new OkHttpClient.Builder();
    if (credentials != null) {
      okHttpClientBuilder.addInterceptor(chain -> chain.proceed(chain.request().newBuilder()
          .header("Authorization", credentials.basic())
          .build()));
    }
    var okHttpClient = okHttpClientBuilder.build();

    return builder()
        .okHttpClient(okHttpClient)
        .serverUrl(streamRegistryUrl)
        .addCustomTypeAdapter(OBJECTNODE, new ObjectNodeTypeAdapter())
        .build();
```

state/graphql-sender/src/main/java/com/expediagroup/streamplatform/streamregistry/state/graphql/Credentials.java
rshotels: Except that it then makes it more complex to confirm that 'credentials' were provided

rshotels: I'm ok with there being a 'credentials' object as a container for name and pass, however I don't think this should be responsible for BASIC encoding. 

Nothing about name/pass credentials implies this encoding, thats decided on by the transport.

state/graphql-sender/src/main/java/com/expediagroup/streamplatform/streamregistry/state/graphql/DefaultApolloClientFactory.java
rshotels: I would suggest something like the above too. There is no need for an 'else' branch in the construction this way. Either we inject the Interceptor or we don't.

state/graphql-sender/src/main/java/com/expediagroup/streamplatform/streamregistry/state/graphql/Credentials.java
nahguam: Yea, that thought went through my head too, but abandoned it due to having to deal with 2 classes named `Credentials`. :D

But yes, a single class with non-nullable fields as a container would be preferable.

state/graphql-sender/src/main/java/com/expediagroup/streamplatform/streamregistry/state/graphql/DefaultApolloClientFactory.java
a-nersharma: done

state/graphql-sender/src/main/java/com/expediagroup/streamplatform/streamregistry/state/graphql/Credentials.java
a-nersharma: done

state/graphql-sender/src/main/java/com/expediagroup/streamplatform/streamregistry/state/graphql/DefaultApolloClientFactory.java
a-nersharma: done

state/graphql-sender/src/main/java/com/expediagroup/streamplatform/streamregistry/state/graphql/DefaultApolloClientFactory.java
```diff
@@ -28,25 +29,34 @@
...
   @Override
   public ApolloClient create() {
-    var builder = builder()
-        .okHttpClient(new OkHttpClient.Builder().build());
-    configurer.accept(builder);
-    return builder
+    var okHttpClientBuilder = new OkHttpClient.Builder();
+    if (credentials != null) {
+      okHttpClientBuilder.addInterceptor(chain -> chain.proceed(chain.request().newBuilder()
+          .header("Authorization", (Base64.getEncoder().encodeToString((credentials.getUserName() + ":" + credentials.getPassword()).getBytes(UTF_8))))
```
rshotels, line 44: Doesn't need the braces around (Base64.....)

rshotels: Are we sure this is the expected "Authorization" value. I think the canonical form is
"Basic <thebytes>", eg: for 'user' and 'pass', the Authorization would be "Basic dXNlcjpwYXNz"

I could be wrong, but would be good to get this confirmed and documented.

rshotels: Perhaps extract to a BasicAuthEncoder and add an illustrative test.

a-nersharma: yes, it needs to be "Basic dXNlcjpwYXNz". will fix that

a-nersharma: fixed

state/graphql-sender/src/main/java/com/expediagroup/streamplatform/streamregistry/state/graphql/DefaultApolloClientFactory.java
```diff
@@ -28,25 +29,34 @@
...
   @Override
   public ApolloClient create() {
-    var builder = builder()
-        .okHttpClient(new OkHttpClient.Builder().build());
-    configurer.accept(builder);
-    return builder
+    var okHttpClientBuilder = new OkHttpClient.Builder();
+    if (credentials != null) {
+      okHttpClientBuilder.addInterceptor(chain -> chain.proceed(chain.request().newBuilder()
+          .header("Authorization", "Basic " + Base64.getEncoder().encodeToString((credentials.getUserName() + ":" + credentials.getPassword()).getBytes(UTF_8)))
```
rshotels, line 44: I'm not seeing any evidence that this is actually accepted by the server.
Perhaps that isn't needed, is the server handling this yet?

Approved with this reservation.

nahguam, line 44: See: `okhttp3.Credentials`. Becomes something like:

```
.header("Authorization", Credentials.basic(c.getUserName(), c.getPassword()))
```

state/graphql-sender/src/main/java/com/expediagroup/streamplatform/streamregistry/state/graphql/DefaultApolloClientFactory.java
```diff
@@ -28,25 +29,34 @@
...
+    }
+
+    var okHttpClient = okHttpClientBuilder.build();
+
+    var builder = builder().okHttpClient(okHttpClient);
+    ApolloClient apolloClient = builder
         .serverUrl(streamRegistryUrl)
         .addCustomTypeAdapter(OBJECTNODE, new ObjectNodeTypeAdapter())
         .build();
+    return apolloClient;
```
nahguam, line 55:
```suggestion
    return builder()
        .okHttpClient(okHttpClient)
        .serverUrl(streamRegistryUrl)
        .addCustomTypeAdapter(OBJECTNODE, new ObjectNodeTypeAdapter())
        .build();
```

state/graphql-sender/src/main/java/com/expediagroup/streamplatform/streamregistry/state/graphql/Credentials.java
```diff
@@ -0,0 +1,33 @@
...
+  @NonNull private final String password;
+
+  public String getUserName() {
+    return userName;
+  }
+
+  public String getPassword() {
+    return password;
+  }
+}
```
nahguam, line 33:
```suggestion
import lombok.NonNull;
import lombok.Value;

@Value
public class Credentials {
  @NonNull String userName;
  @NonNull String password;
}
```

state/graphql-sender/src/main/java/com/expediagroup/streamplatform/streamregistry/state/graphql/DefaultApolloClientFactory.java
a-nersharma: done

state/graphql-sender/src/main/java/com/expediagroup/streamplatform/streamregistry/state/graphql/DefaultApolloClientFactory.java
a-nersharma: done

state/graphql-sender/src/main/java/com/expediagroup/streamplatform/streamregistry/state/graphql/Credentials.java
a-nersharma: done

PR #259, opened 2020-09-18
Add CLI command to discover and delete all child entities of a stream.

nahguam:
# stream-registry PR

# PR Checklist Forms

- [ ] CHANGELOG.md updated
- [ ] Reviewer assigned


KasunDon:
@nahguam looks good to me, I don't think i can approve this it's in Draft status

nahguam:
I agree with @teabot's suggestion above. This was the long term plan, but the intent of the first iteration was to get something out quickly. I've changed the CLI to be more in line with the long term plan.
i.e. changed `streamDeleteCascade` to `delete stream --cascade`. Currently stream is the only entity supporting the `--cascade` flag.

cli/src/main/java/com/expediagroup/streamplatform/streamregistry/cli/command/delete/Delete.java
```diff
@@ -0,0 +1,221 @@
...
+      return unmodifiableList(client.schemas(domain, schema));
+    }
+  }
+
+  @Command(name = "stream")
+  static class Stream extends Base {
+    @Option(names = "--domain", required = true) private String domain;
+    @Option(names = "--stream", required = true) private String stream;
+    @Option(names = "--version", required = true) private int version;
+    @Option(names = "--cascade") boolean cascade;
```
irinack, line 106: Should this be private?

cli/src/main/java/com/expediagroup/streamplatform/streamregistry/cli/command/delete/EntityClient.java
```diff
@@ -0,0 +1,202 @@
...
+  private final ApolloExecutor executor;
+  private final KeyConverter converter;
+
+  @Value
+  static class StreamKeyWithSchemaKey {
+    StreamKey stream;
+    SchemaKey schema;
+  }
+
+  List<DomainKey> domains(String domain) {
```
irinack, line 66: This is an optional change, because I'm being slightly pedantic, but I think a more meaningful method name would help clear up what this method does. I like using verbs so changing this to something like "getDomains" or "getDomainKeys" would be better imo. (Same applies to the rest of the getters in this class)

For example, earlier in the code I was trying to figure out what this line does 
`return unmodifiableList(client.domains(domain))` and it was slightly confusing to me until I reached this point where the method is defined.

cli/src/main/java/com/expediagroup/streamplatform/streamregistry/cli/command/delete/Delete.java
nahguam: yes!

cli/src/main/java/com/expediagroup/streamplatform/streamregistry/cli/command/delete/Formatter.java
```diff
@@ -0,0 +1,214 @@
...
+  private static abstract class NestedKeySerializer<K> extends StdSerializer<K> {
+    private final String name;
+
+    protected NestedKeySerializer(Class<K> kClass, String name) {
+      super(kClass);
+      this.name = name;
+    }
+
+    @Override
+    public void serialize(K key, JsonGenerator g, SerializerProvider p) throws IOException {
```
irinack, line 75: SerializerProvider p is not being used here, right?

```suggestion
    public void serialize(K key, JsonGenerator jsonGenerator) throws IOException {
```

cli/src/main/java/com/expediagroup/streamplatform/streamregistry/cli/command/delete/EntityClient.java
nahguam: I generally prefer brevity and think `get` is redundant. I agree it should specify it's getting the keys. Happy to change if it makes it easier to understand.

cli/src/main/java/com/expediagroup/streamplatform/streamregistry/cli/command/delete/Formatter.java
nahguam: This method signature is from `StdSerializer`. The `void serialize(K key, JsonGenerator g)` signature it the one I've defined in `NestedKeySerializer`

cli/src/test/java/com/expediagroup/streamplatform/streamregistry/cli/command/delete/DeleteTest.java
```diff
@@ -0,0 +1,333 @@
...
+import com.expediagroup.streamplatform.streamregistry.state.model.Entity.StreamKey;
+import com.expediagroup.streamplatform.streamregistry.state.model.Entity.ZoneKey;
+
+@RunWith(MockitoJUnitRunner.Silent.class)
+public class DeleteTest {
+  private final CommandLine underTest = new CommandLine(new Delete());
+
+  private final String[] standardOptions = {"--bootstrapServers=bootstrapServers", "--schemaRegistryUrl=schemaRegistryUrl", "--streamRegistryUrl=streamRegistryUrl"};
+
+  private final DomainKey domainKey = new DomainKey("domain");
```
irinack, line 53: This is optional, but since you're using the strings "domain", "schema", "stream", "producer" and "consumer" in several places, it would have been easier to declare them as static final fields instead of hardcoding it :D

irinack (approved):

cli/src/test/java/com/expediagroup/streamplatform/streamregistry/cli/command/delete/DeleteTest.java
nahguam: done

fantayeneh (approved):
LGTM

PR #265, opened 2020-11-02
refactored upserts

a-nersharma:
# stream-registry PR

Access control on various stream-registry entities

### Added
- Added PreAuthorize for get, find, and findAll
- Added get method in services for all entities and applied PreAuthorize on the get method

### Changed
- Updated upsert methods in mutations and removed upsert methods from services
- Refactored updateStatus

# PR Checklist Forms

- [x] CHANGELOG.md updated
- [x] Reviewer assigned


teabot (approved):

teabot (dismissed):
This looks ok, I guess the Expression handler will be where it all happens?

a-nersharma:
> This looks ok, I guess the Expression handler will be where it all happens?

Yes. In another PR I have the expressionhandler implementation

core/src/main/java/com/expediagroup/streamplatform/streamregistry/core/services/ConsumerBindingService.java
```diff
@@ -64,9 +66,8 @@
     return save(consumerBinding);
   }
 
-  @PreAuthorize("hasPermission(#consumerBindingKey, 'UPDATE_STATUS')")
-  public Optional<ConsumerBinding> updateStatus(ConsumerBindingKey consumerBindingKey, Status status) {
-    ConsumerBinding consumerBinding = read(consumerBindingKey).get();
+  @PreAuthorize("hasPermission(#consumerBinding, 'UPDATE_STATUS')")
+  public Optional<ConsumerBinding> updateStatus(ConsumerBinding consumerBinding, Status status) {
```
nahguam, line 70: Updating the binding status shouldn't need the whole binding?

core/src/main/java/com/expediagroup/streamplatform/streamregistry/core/services/ZoneService.java
```diff
@@ -74,14 +75,16 @@
     return Optional.ofNullable(zone);
   }
 
-  public Optional<Zone> upsert(Zone zone) throws ValidationException {
-    return !read(zone.getKey()).isPresent() ? create(zone) : update(zone);
+  @PostAuthorize("returnObject.isEmpty() ? true: hasPermission(returnObject, 'READ')")
+  public Optional<Zone> get(ZoneKey key) {
```
nahguam, line 79: Not sure how I feel about maintaining both `get` (authz) and `find` (no-authz) methods. If it's specifically to bypass authz then I think they should be named as such so it's clear. However, I think it's ok to only have one method (`get`) on the basis that you need to be able to READ an entity to also be able to UPDATE it.

core/src/main/java/com/expediagroup/streamplatform/streamregistry/core/services/ConsumerBindingService.java
a-nersharma: The RangerAuthoriser contract needs the full object (Entity<*>).
This logic has been pushed up to the graphql layer?

core/src/main/java/com/expediagroup/streamplatform/streamregistry/core/services/ZoneService.java
a-nersharma: I will rename the methods appropriately.

If we maintain only one method (for get and read), then RangerAuthoriser will get two calls for any update or update_status (one for get and one for update). Also validators will result in further RangerAuthoriser calls.

nahguam (dismissed):

PR #270, opened 2021-01-19
Support recovering deletes that have been issued while an Agent is offline

jamespfaulkner:
# stream-registry PR

Currently deletes that are issued while an agent is offline are simply missed, without an inbuilt way of finding out what they were. We now keep a record of these in the backing state which an agent can process and purge when it comes back online.

### Added
* `EntityView::allDeleted` added to facilitate the reading of all deleted Entities
* `EntityView::purgeDeleted` added to facilitate the removal of deleted Entities that have been handled

# PR Checklist Forms

- [ ] CHANGELOG.md updated
- [ ] Reviewer assigned


state/core/src/main/java/com/expediagroup/streamplatform/streamregistry/state/StateKey.java
```diff
@@ -0,0 +1,37 @@
...
+import lombok.AllArgsConstructor;
+import lombok.EqualsAndHashCode;
+import lombok.ToString;
+
+import com.expediagroup.streamplatform.streamregistry.state.model.Entity;
+
+@EqualsAndHashCode
+@ToString
+@AllArgsConstructor
+class StateKey {
```
nahguam, line 27: I had assumed the `deleted` flag would have been added to `Entity`, not creating a new wrapper key. I can't actually think of any reason why this is bad though. It does open us up to the potential bug of having both the deleted and undeleted key present at the same time though.

state/api/src/main/java/com/expediagroup/streamplatform/streamregistry/state/EntityView.java
```diff
@@ -63,4 +63,25 @@
...
+
+  /**
+   * Purge a deleted entity so that it no longer appears in {@link #allDeleted(Class)}.
+   *
+   * @param key the key of the entity to be purged
+   * @param <K> the key type.
+   * @param <S> the specification type.
+   * @return an optional containing the entity that was purged or empty if no entity exists.
+   */
+  <K extends Key<S>, S extends Specification> Optional<Entity<K, S>> purgeDeleted(K key);
```
nahguam, line 86: Stream Registry itself uses this API in the repository module and it currently has no need to react to events. Now it does - to purge deleted items from the view.

The `entityView` bean in `KafkaConfiguration` in the `repository-kafka` module will need an `EntityViewListener` implementation that calls `purgeDeleted` for `SpecificationDeletionEvent`s. 

The `EntityView#load()` can also probably be removed.

state/core/src/main/java/com/expediagroup/streamplatform/streamregistry/state/EntityViewUpdater.java
```diff
@@ -51,44 +54,51 @@
...
     val entity = new Entity<>(event.getKey(), oldEntity.getSpecification(), oldEntity.getStatus().with(event.getStatusEntry()));
-    entities.put(event.getKey(), entity);
+    entities.put(existing(event.getKey()), entity);
     log.debug("Updated {} with {}", event.getKey(), event.getStatusEntry());
     return oldEntity;
   }
 
   private <K extends Entity.Key<S>, S extends Specification> Entity<K, S> delete(SpecificationDeletionEvent<K, S> event) {
-    val oldEntity = (Entity<K, S>) entities.remove(event.getKey());
+    val oldEntity = (Entity<K, S>) entities.remove(existing(event.getKey()));
```
nahguam, line 88: Yea, here - if the `deleted` flag was on `Entity` then you wouldn't need to remove/add, you'd just replace/mutate the existing entity with the deleted version.

state/core/src/main/java/com/expediagroup/streamplatform/streamregistry/state/StateKey.java
jamespfaulkner: I was trying to cover the case where while an Agent is offline, a delete is fired followed by another create using the same key. 

I didn't want an offline agent to miss that delete, but the more I'm playing with this implementation in an actual agent the more I'm realising that:
1. it's difficult for the agent to know what to do in that situation
2. the juice isn't worth the squeeze for handling what is probably going to be an extreme corner case.

state/api/src/main/java/com/expediagroup/streamplatform/streamregistry/state/EntityView.java
jamespfaulkner: Great, that makes sense. I wasn't too sure how was best to handle the purge in Stream Registry.

state/core/src/main/java/com/expediagroup/streamplatform/streamregistry/state/EntityViewUpdater.java
jamespfaulkner: Coolios, makes sense.

state/api/src/main/java/com/expediagroup/streamplatform/streamregistry/state/EntityView.java
```diff
@@ -36,13 +37,6 @@
    */
   CompletableFuture<Void> load(EntityViewListener listener);
 
-  /**
-   * Commences loading the view. The returned {@link CompletableFuture} completes when the view has fully loaded.
-   *
-   * @return a future that completes when the view has fully loaded.
-   */
-  CompletableFuture<Void> load();
-
```
irinack, line 45: Are we sure this isn't being used by any customers? This is a part of the library that is accessible to them, right?

irinack: I noticed @nahguam's comment below so maybe he can answer this

state/core/src/main/java/com/expediagroup/streamplatform/streamregistry/state/DefaultEntityView.java
```diff
@@ -54,21 +55,37 @@ public DefaultEntityView(EventReceiver receiver) {
...
+      .filter(it -> it.entity.getKey().getClass().equals(keyClass))
+      .map(it -> (Entity<K, S>) it.entity);
+  }
+
+  @Override
+  public <K extends Entity.Key<S>, S extends Specification> Map<K, Optional<Entity<K, S>>> allDeleted(Class<K> keyClass) {
+//    return entities.entrySet().stream()
+//      .filter(it -> it.getValue().deleted)
+//      .filter(it -> it.getKey().getClass().equals(keyClass))
+//      .map(it -> (K) it.getKey());
```
irinack, line 76: Did you mean to commit this?

state/api/src/main/java/com/expediagroup/streamplatform/streamregistry/state/EntityView.java
jamespfaulkner: > The EntityView#load() can also probably be removed.

This was only used in `KafkaConfiguration`. None of our clients used it. If anyone else was using it, I think it makes sense for them to actually explicitly define which `EntityViewListener` they want to use.

state/core/src/main/java/com/expediagroup/streamplatform/streamregistry/state/DefaultEntityView.java
jamespfaulkner: D'oh. Nope, will remove

state/core/src/main/java/com/expediagroup/streamplatform/streamregistry/state/EntityViewUpdater.java
```diff
@@ -51,45 +55,62 @@
     }
   }
 
+  <K extends Entity.Key<S>, S extends Specification> Optional<Entity<K, S>> purge(K key) {
+    val stateEntity = Optional.ofNullable(entities.get(key))
+      .filter(it -> it.deleted);
+    stateEntity.ifPresent(it -> {
+      entities.remove(key);
+      log.debug("Purged entity for {}", key);
```
irinack, line 63: This is minor, but it's more Splunk friendly if you do something like 
```suggestion
      log.debug("Purged entity for key={}", key);
```

irinack: Although I did notice that we don't tend to log key value pairs :'(

state/core/src/test/java/com/expediagroup/streamplatform/streamregistry/state/EntityViewUpdaterTest.java
```diff
@@ -18,68 +18,194 @@
...
+    assertThat(entities, hasEntry(key, deleted(null)));
+  }
+
+  @Test
+  public void deleteSpecificationEvent() {
+    val previousEntity = underTest.update(specificationEvent);
+
+    assertThat(previousEntity, is(nullValue()));
+    assertThat(entities.get(key), is(existing(entity.withStatus(oldStatus))));
+    assertThat(entities, hasEntry(key, existing(entity.withStatus(oldStatus))));
```
irinack, line 132: Aren't these 2 checking for the same thing?

state/core/src/test/java/com/expediagroup/streamplatform/streamregistry/state/EntityViewUpdaterTest.java
```diff
@@ -18,68 +18,194 @@
...
+    assertThat(entities, is(aMapWithSize(1)));
```
irinack, line 133: This assertion should probably go first on line 131. It's not worth asserting the rest if this fails

state/core/src/test/java/com/expediagroup/streamplatform/streamregistry/state/EntityViewUpdaterTest.java
```diff
@@ -18,68 +18,194 @@
...
+
+    val deletedEntity = underTest.update(specificationDeletionEvent);
+    assertThat(deletedEntity, is(entity.withStatus(oldStatus)));
+    assertThat(entities.get(key), is(deleted(entity.withStatus(oldStatus))));
+    assertThat(entities, is(aMapWithSize(1)));
```
irinack, line 138: Same here. I would switch lines 137 and 138

state/core/src/main/java/com/expediagroup/streamplatform/streamregistry/state/EntityViewUpdater.java
jamespfaulkner: done

state/core/src/test/java/com/expediagroup/streamplatform/streamregistry/state/EntityViewUpdaterTest.java
```diff
@@ -18,68 +18,191 @@
...
+    assertThat(entities, hasEntry(key, deleted(null)));
+
+    underTest.purge(key);
+
+    assertThat(entities, is(aMapWithSize(0)));
+  }
+
+  @Test(expected = IllegalArgumentException.class)
+  public void unhandledEvents() {
```
irinack, line 205: This is minor but the method name is quite vague. 
```suggestion
  public void updateWithNull() {
```

state/it/src/test/java/com/expediagroup/streamplatform/streamregistry/state/it/AgentIT.java
```diff
@@ -0,0 +1,276 @@
...
+
+  @Rule
+  public KafkaContainer kafka = new KafkaContainer();
+
+  @Test
+  public void testBootstrapping() {
+    val topicName = topicName();
+    val kafkaEventSender = kafkaEventSender(topicName);
+    val entityView = new DefaultEntityView(kafkaEventReceiver(topicName, "groupId"));
+    val dummyAgent = new StoringEntityViewListener();
```
irinack, line 79: I think you could add this to a BeforeEach? These lines occur in every test right?

irinack (approved):

repository/kafka/src/main/java/com/expediagroup/streamplatform/streamregistry/repository/kafka/PurgingEntityViewListener.java
```diff
@@ -0,0 +1,55 @@
...
+import com.expediagroup.streamplatform.streamregistry.state.model.specification.Specification;
+
+@AllArgsConstructor
+class PurgingEntityViewListener implements EntityViewListener {
+  private final EntityView entityView;
+
+  @Override
+  public <K extends Entity.Key<S>, S extends Specification> void onEvent(Entity<K, S> oldEntity, @NonNull Event<K, S> event) {
+    if (event instanceof SpecificationDeletionEvent) {
+      entityView.purgeDeleted(event.getKey());
```
techabstraction, line 35: Is this going to cause everything to be purged before an offline agent has a chance to see it?

jamespfaulkner: Yeah, thats the idea 😄 

This is currently only used by the `stream-registry` which has no use for knowing about deletes. See The conversation with @nahguam above (https://github.com/ExpediaGroup/stream-registry/pull/270#discussion_r561713684)

techabstraction: Thanks for the offline discussion @jamespfaulkner. For the benefit of others, each agent has it's own copy of the entity view so this purge will not affect what they see.

state/api/src/main/java/com/expediagroup/streamplatform/streamregistry/state/EntityView.java
```diff
@@ -63,4 +57,25 @@
...
+   */
+  <K extends Key<S>, S extends Specification> Map<K, Optional<Entity<K, S>>> allDeleted(Class<K> keyClass);
+
+  /**
+   * Purge a deleted entity so that it no longer appears in {@link #allDeleted(Class)}.
+   *
+   * @param key the key of the entity to be purged
+   * @param <K> the key type.
+   * @param <S> the specification type.
+   * @return an optional containing the entity that was purged or empty if no entity exists.
```
teabot, line 78: Minor point here, but can we in the javadoc of this contract define "purge" and "delete" and perhaps have a few sentences on the lifecycle of deletes wrt to both stream registry, and more generally: agents.

repository/kafka/src/main/java/com/expediagroup/streamplatform/streamregistry/repository/kafka/KafkaConfiguration.java
```diff
@@ -37,40 +37,43 @@ EventCorrelator eventCorrelator() {
 
   @Bean
   EventSender eventSender(
-      @Value("${repository.kafka.bootstrapServers}") String bootstrapServers,
```
nahguam, line 40: Can we avoid the formatting flip-flip :)

repository/kafka/src/main/java/com/expediagroup/streamplatform/streamregistry/repository/kafka/PurgingEntityViewListener.java
nahguam: Remember this is only in-memory local to an app, this doesn't affect other apps. They will have their own in-memory view of the logically deleted (tombstoned) entities.

state/api/src/main/java/com/expediagroup/streamplatform/streamregistry/state/EntityView.java
```diff
@@ -63,4 +57,25 @@
...
+  /**
+   * Returns a {@link Map} containing all keys of the given {@link Key} type which have been deleted but not
+   * purged ({@link #purgeDeleted(Key)}) mapped to the deleted Entity (if known)
+   *
+   * @param keyClass the key class of an entity type.
+   * @param <K>      the key type.
+   * @param <S>      the specification type.
+   * @return a map containing all deleted keys of a give type to the key if known
+   */
+  <K extends Key<S>, S extends Specification> Map<K, Optional<Entity<K, S>>> allDeleted(Class<K> keyClass);
```
teabot, line 70: I recall that this `Optional` is required to handle possible edge case. It's be great to have one or two sentences in the javadoc to explain that situation so that callers can have an idea of what they might get back and when.

repository/kafka/src/main/java/com/expediagroup/streamplatform/streamregistry/repository/kafka/KafkaConfiguration.java
```diff
@@ -37,40 +37,43 @@ EventCorrelator eventCorrelator() {
...
+      .schemaRegistryUrl(schemaRegistryUrl)
+      .build();
     return new KafkaEventReceiver(receiverConfig, eventCorrelator);
   }
 
   @Bean
   EntityView entityView(EventReceiver eventReceiver) {
     EntityView entityView = new DefaultEntityView(eventReceiver);
-    entityView.load().join();
+    PurgingEntityViewListener entityViewListener = new PurgingEntityViewListener(entityView);
```
nahguam, line 73: Two way dependency `PurgingEntityViewListener` <-> `EntityView` but I don't think we can avoid that.

state/api/src/main/java/com/expediagroup/streamplatform/streamregistry/state/EntityView.java
nahguam: resolved with `PurgingEntityViewListener`

state/core/src/main/java/com/expediagroup/streamplatform/streamregistry/state/StateValue.java
```diff
@@ -0,0 +1,37 @@
...
+import lombok.AllArgsConstructor;
+import lombok.EqualsAndHashCode;
+import lombok.ToString;
+
+import com.expediagroup.streamplatform.streamregistry.state.model.Entity;
+
+@EqualsAndHashCode
+@ToString
+@AllArgsConstructor
+public class StateValue {
```
nahguam, line 27: Is it not simpler to just add `boolean deleted` directly onto `Entity` (genuine question)?

teabot (changes requested):
This feels complete, but (necessarily) complex conceptually. Given that this contract is the key integration point for our agents, and indeed other external integrations, I believe it would be beneficial to invest a little time on javadoc that explains the concepts, terms, and sets expectations on the operation of methods in the **contract**. I do not think there can be too much in this regard, perhaps even a readme with a plantuml sequence diagram or some such (different PR for that one is fine).

repository/kafka/src/main/java/com/expediagroup/streamplatform/streamregistry/repository/kafka/KafkaConfiguration.java
jamespfaulkner: Should we be using 2 or 4 space indentation?

I'm happy to change back but I think this indentation is now consistent with the formatting in other classes.

repository/kafka/src/main/java/com/expediagroup/streamplatform/streamregistry/repository/kafka/KafkaConfiguration.java
jamespfaulkner: Yeah, agreed this is a bit pants. I originally tried to hide this by putting the purging in the constructor of `PurgingEntityViewListener` but I think that looked worse.

state/core/src/main/java/com/expediagroup/streamplatform/streamregistry/state/EntityViewUpdater.java
```diff
@@ -51,45 +55,62 @@
...
   private <K extends Entity.Key<S>, S extends Specification> Entity<K, S> delete(StatusDeletionEvent<K, S> event) {
-    val oldEntity = (Entity<K, S>) entities.get(event.getKey());
+    val oldEntity = (Entity<K, S>) getExistingEntity(event.getKey());
     if (oldEntity == null) {
       log.warn("Received status deletion {} for non existent entity {}", event.getStatusName(), event.getKey());
       return null;
     }
     val entity = oldEntity.withStatus(oldEntity.getStatus().without(event.getStatusName()));
-    entities.put(event.getKey(), entity);
+    entities.put(event.getKey(), existing(entity));
```
techabstraction, line 105: Should it be this?
```suggestion
    entities.put(event.getKey(), deleted(entity));
```

state/api/src/main/java/com/expediagroup/streamplatform/streamregistry/state/EntityView.java
jamespfaulkner: Yeah, I'll try and articulate the case just so I don't forget 😄 

Say we have the following on a backing kafka topic:

| Key | Value |
|---|---|
| `K1` | `V1` |

After bootstrapping the backing Map contains `K1` -> `(K2, deleted=false)`.

We then issue a delete:

| Key | Value |
|---|---|
| `K1` | `V1` |
| `K1` | `null` |

After bootstrapping the backing Map contains `K1` -> `(K2, deleted=true)`.

Sometime in the future, compaction will run. But this does not cause tombstones to be deleted right away. (This is by design in Kafka, in order to give consumers time to register the delete). So we end up with:

| Key | Value |
|---|---|
| `K1` | `null` |

Now we no longer have the original entity. So the backing map looks like:

`K1` -> `(null, deleted=true)`.

I consumed from a Kafka topic in usage and could confirm this behaviour.

jamespfaulkner: The above being said, you are correct that the contract needs more explanation in these java docs 😄

repository/kafka/src/main/java/com/expediagroup/streamplatform/streamregistry/repository/kafka/KafkaConfiguration.java
techabstraction: We could do a separate pr just to fix formatting if it affects many classes? Perhaps adding .editorconfig at same time.

state/core/src/main/java/com/expediagroup/streamplatform/streamregistry/state/EntityViewUpdater.java
jamespfaulkner: Eagle eyed 👍 

So hopefully there are test cases to cover this, but no I don't believe this should be. 

In this method we are deleting the Status rather than deleting the Entity. So we want to remove the Status from an existing Entity, but not actually delete that Entity.

Another minor point, it's not possible to delete a status on a delete entity, which I believe is the correct functionality.

state/core/src/main/java/com/expediagroup/streamplatform/streamregistry/state/StateValue.java
jamespfaulkner: So there was a couple of reasons I didn't go for that approach:

1. We'd then expose another way of checking/retrieving deleted `Entity`s. Since we made this backwards compatible, the `EntityView::all` only returns `Entities` that aren't deleted. If there was a `deleted` flag on entity it might encourage users to query that value when its not needed.
2. When filter for deleted Entities, with this approach we only need to check the `deleted` flag on `StateValue`. If we added the flag to the entity there are two states where an entity is deleted (when `Entity::deleted == true` and when the entity itself is `null`).
3. I'm lazy 😛 . Since `Entity` is an interface, we'd need to implement `deleted` in all implementations of it. Obviously we could add something to the class hierarchy to handle this for us, but that just feels like more complication.

Does that make sense?

state/core/src/main/java/com/expediagroup/streamplatform/streamregistry/state/StateKey.java
teabot: > juice isn't worth the squeeze

love it

state/api/src/main/java/com/expediagroup/streamplatform/streamregistry/state/EntityView.java
jamespfaulkner: Done, though I've not included any references to the specific kafka case since this class doesn't have requirements on the underlying storage mechanism.

state/api/src/main/java/com/expediagroup/streamplatform/streamregistry/state/EntityView.java
jamespfaulkner: I've tried to explain the concepts/lifecycle in a bit more depth. I think you are correct that a more indepth readme is the way to go with this one. I'll do that in a followup PR.

teabot (approved):

state/api/src/main/java/com/expediagroup/streamplatform/streamregistry/state/EntityView.java
```diff
@@ -63,4 +76,30 @@
...
    */
   <K extends Key<S>, S extends Specification> Stream<Entity<K, S>> all(Class<K> keyClass);
+
+  /**
+   * Returns a {@link Map} containing all keys of the given {@link Key} type which have been deleted but not
+   * purged ({@link #purgeDeleted(Key)}) mapped to the deleted Entity (if known).
+   * <p/>
+   * Depending on underlying storage it is possible for an Entity to be marked deleted, by all references to the
+   * original entity having removed. In that case we still know that an entity with a given key has been deleted
+   * and the map will contain an {@link Map.Entry} with an empty value.
```
techabstraction, line 86:
```suggestion
   * Depending on underlying storage it is possible for an Entity to be marked deleted, but all references to the
   * original entity have been removed. In that case we still know that an entity with a given key has been deleted
   * and the map will contain an {@link Map.Entry} with an empty value.
```

state/api/src/main/java/com/expediagroup/streamplatform/streamregistry/state/EntityView.java
```diff
@@ -63,4 +76,30 @@
...
+   *
+   * @param keyClass the key class of an entity type.
+   * @param <K>      the key type.
+   * @param <S>      the specification type.
+   * @return a map containing all deleted keys of a give type to the previously existing entity (if known)
```
techabstraction, line 91:
```suggestion
   * @return a map containing all deleted keys of a given type to the previously existing entities (if known)
```

state/api/src/main/java/com/expediagroup/streamplatform/streamregistry/state/EntityView.java
```diff
@@ -25,6 +26,25 @@
 
 /**
  * Provides a unified view of the current state of all entities held in Stream Registry.
+ * <p/>
+ * The underlying state storage from which this is this is loaded may support entities which have been deleted. If
+ * that is the case it is important to understand the lifecycle of entities as it relates to {@link EntityView}.
+ * <p/>
+ * Entities may be created and updated in the underlying state storage. Any entity which has been created (but not
+ * deleted) will be returned by {@link #all(Class)} and {@link #get(Key)}. It is expected that users handle said
+ * entities accordingly (update external dependencies/respond to view requests etc).
```
techabstraction, line 35: Perhaps simplify it:
```suggestion
 * Entities may be created and updated in the underlying state storage. Any entity which has been created (but not
 * deleted) will be returned by {@link #all(Class)} and {@link #get(Key)}. 
```

state/api/src/main/java/com/expediagroup/streamplatform/streamregistry/state/EntityView.java
```diff
@@ -25,6 +26,25 @@
...
+ * {@link #get(Key)}. Instead they will be returned by {@link #allDeleted(Class)}. In this case it is expected that
+ * users handle those delete requests accordingly (e.g. remove the entity from external dependencies). This handling
+ * does not automatically remove the entities from this in memory state (or the underlying state storage). In order to
+ * remove the entity, {@link #purgeDeleted(Key)} should be used. This will remove all in memory references to said
+ * entity and it will no longer be returned by {@link #allDeleted(Class)}.
+ * <p/>
+ * <i>Note:</i> {@link #purgeDeleted(Key)} does not delete the entity from the underlying storage, only from memory. It
+ * is therefore possible that an entity that has previously had it's delete handled, will appear in
+ * {@link #allDeleted(Class)} after restarts. Users of this class should be aware and handle this case (e.g. by
+ * simply calling {@link #purgeDeleted(Key)} on said entities)
```
techabstraction, line 47:
```suggestion
 * simply calling {@link #purgeDeleted(Key)} on such entities)
```

state/api/src/main/java/com/expediagroup/streamplatform/streamregistry/state/EntityView.java
```diff
@@ -25,6 +26,25 @@
...
+ * Entities may be created and updated in the underlying state storage. Any entity which has been created (but not
+ * deleted) will be returned by {@link #all(Class)} and {@link #get(Key)}. It is expected that users handle said
+ * entities accordingly (update external dependencies/respond to view requests etc).
+ * <p/>
+ * When entities are deleted in the underlying state storage, they will stop being returned by {@link #all(Class)} and
+ * {@link #get(Key)}. Instead they will be returned by {@link #allDeleted(Class)}. In this case it is expected that
+ * users handle those delete requests accordingly (e.g. remove the entity from external dependencies). This handling
+ * does not automatically remove the entities from this in memory state (or the underlying state storage). In order to
+ * remove the entity, {@link #purgeDeleted(Key)} should be used. This will remove all in memory references to said
+ * entity and it will no longer be returned by {@link #allDeleted(Class)}.
```
techabstraction, line 42: Line length/wrapping needs fixing, but here's a suggestion for the wording:
```suggestion
 * When entities are deleted in the underlying state storage, they will stop being returned by {@link #all(Class)} and
 * {@link #get(Key)}. Instead they will be returned by {@link #allDeleted(Class)}. Agents can use this method to find deleted entities after an offline period (restarts or maintenance) where they may have missed the deletion event while offline. They can then handle those deleted entities accordingly (e.g. remove the entity from external dependencies etc). This handling
 * does not automatically remove the entities from this in memory state (or the underlying state storage). In order to
 * remove the entity, {@link #purgeDeleted(Key)} should be used. This will remove all in memory references to the
 * deleted entity and it will no longer be returned by {@link #allDeleted(Class)}.
```

techabstraction (approved):

nahguam (approved):

PR #271, opened 2021-01-20
Downgrade to Java 8

nahguam:


state/graphql-sender/src/test/java/com/expediagroup/streamplatform/streamregistry/state/graphql/ApolloExecutorTest.java
```diff
@@ -67,12 +71,12 @@ public void onResponseNoErrors() {
   public void onResponseErrors() {
     when(client.mutate(mutation)).thenReturn(mutationCall);
 
-    var result = underTest.execute(mutation);
+    CompletableFuture<? extends Response<?>> result = underTest.execute(mutation);
```
irinack, line 74: Why not just use val like the others?

irinack (approved):

state/graphql-sender/src/test/java/com/expediagroup/streamplatform/streamregistry/state/graphql/ApolloExecutorTest.java
nahguam: The compiler couldn't determine the generic type. As with most of these cases, mash `?` `extends` `super` until it works 😆

jamespfaulkner (approved):

PR #283, opened 2021-04-14
STREAMP-3923 securityspec

andykruth-ii:
Adding a `security` block to Entity Specifications.
* Entity Model gets a security type of Map<Role, List<Principal>>
* State Model (Avro) only allows string keys in maps so its type is Map<String, List<Principal>>
* GraphQL doesn't like Maps at all, so it uses a strongly typed List[Role(role: string, principals: List<string>)]
* converters in place to move between these types. 

# Added
Added security to Specification models.

# Changed
Added security to AvroSpecification and AvroStreamSpecification schemas, defaulting to empty map.


model/src/main/java/com/expediagroup/streamplatform/streamregistry/model/Principal.java
```diff
@@ -0,0 +1,25 @@
...
+package com.expediagroup.streamplatform.streamregistry.model;
+
+import lombok.AllArgsConstructor;
+import lombok.Data;
+
+@Data
+@AllArgsConstructor
+public class Principal {
+  private String name;
+}
```
nahguam, line 25: Does it need to be mutable? Also non-null `name`?

```suggestion
import lombok.Value;

@Value
public class Principal {
  String name;
}
```

repository/kafka/src/test/java/com/expediagroup/streamplatform/streamregistry/repository/kafka/SampleModel.java
```diff
@@ -55,6 +61,10 @@ static Specification specification() {
     specification.setTags(Collections.singletonList(new Tag("name", "value")));
     specification.setType("type");
     specification.setConfiguration(mapper.createObjectNode());
+    specification.setSecurity(java.util.stream.Stream.of(
```
nahguam, line 64: Kotlin would be nice here :'(

guava might help with the verbosity here

state/graphql-sender/src/main/java/com/expediagroup/streamplatform/streamregistry/state/graphql/GraphQLConverter.java
```diff
@@ -116,6 +119,11 @@ default SpecificationInput convertSpecification(S specification) {
               .collect(toList()))
           .type(specification.getType())
           .configuration(specification.getConfiguration())
+          .security(specification.getSecurity().entrySet().stream().map(
```
nahguam, line 122: Is there a complementary side to this that converts from the input list to a map?

graphql/api/src/main/java/com/expediagroup/streamplatform/streamregistry/graphql/model/inputs/SpecificationInput.java
```diff
@@ -33,6 +36,7 @@
   List<TagInput> tags;
   String type;
   ObjectNode configuration;
+  Map<Role, List<Principal>> security;
```
nahguam, line 39: Does this work defining it as a map here when it's `security: [Role!]!` in the graphql definition?

nahguam (commented):
Big PR but very cleanly presented. Nice work!

graphql/api/src/main/java/com/expediagroup/streamplatform/streamregistry/graphql/model/inputs/SpecificationInput.java
andykruth-ii: no it does not ... Oversight on my part after the change to the graphql schema. Incidentally this also fixes the broken IT.

state/graphql-sender/src/main/java/com/expediagroup/streamplatform/streamregistry/state/graphql/GraphQLConverter.java
andykruth-ii: There is now! See the changes I made to the SpecificationInput class, it has the `asSpecification` method which takes the list of Roles and converts to the Map Role -> List<Principal>

repository/kafka/src/test/java/com/expediagroup/streamplatform/streamregistry/repository/kafka/SampleModel.java
andykruth-ii: yes, guava reduced the complexity, I am replacing all the repetitions.

model/src/main/java/com/expediagroup/streamplatform/streamregistry/model/Principal.java
andykruth-ii: OK - taking this change for both Role and Principal.

cli/src/main/java/com/expediagroup/streamplatform/streamregistry/cli/command/Apply.java
```diff
@@ -71,12 +76,23 @@
     protected String type;
     @Option(names = "--configuration", required = true, converter = ObjectNodeConverter.class)
     protected ObjectNode configuration;
+    @Option(names = "--security", converter = ObjectNodeConverter.class)
+    protected ObjectNode security = new ObjectMapper().createObjectNode();
```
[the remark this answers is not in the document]
andykruth-ii: 👍

repository/kafka/pom.xml
```diff
@@ -46,6 +46,10 @@
       <groupId>org.springframework</groupId>
       <artifactId>spring-context</artifactId>
     </dependency>
+    <dependency>
```
jamespfaulkner, line 49: Minor: Looks like this is only used in tests. Maybe just scope to test?

jamespfaulkner (commented):
LGTM. One minor comment and the `StreamRegistryIT` seems to be broken but other than that 👍

state/graphql-sender/pom.xml
```diff
@@ -55,6 +55,10 @@
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-api</artifactId>
     </dependency>
+    <dependency>
```
nahguam, line 58: graphql-sender is used by the sdk - maybe better not to inflict the transitive dependency on users - or shade&relocate it.

To avoid the guava dep but keep the single-statement approach - another alternative might be something like:

```
new HashMap<String, Principal>() {{
  put("admin", Arrays.asList(new Principal("user1")));
  put("creator", Arrays.asList(new Principal("user2"), new Principal("user3")));
}}
```

graphql/api/src/main/resources/stream-registry.graphql
```diff
@@ -33,24 +33,44 @@ type Tag {
 
 ############ Specification ############
 
+input RoleInput {
```
nahguam, line 36: Would this be better named `SecurityInput`? Same for `Role` -> `Security`? More consistent with the variable name and that fact it is represents a collection of mappings between roles and principals.

graphql/api/src/main/resources/stream-registry.graphql
```diff
@@ -33,24 +33,44 @@ type Tag {
...
+
+input SecurityQuery {
+    roleRegex: String
+    principalRegex: String
+}
+
+type Role {
+    role: String
+    principals: [String]
+}
```
nahguam, line 49:
```suggestion
type Role {
    role: String!
    principals: [String!]!
}
```

nahguam: check the others too?

graphql/api/src/main/resources/stream-registry.graphql
```diff
@@ -33,24 +33,44 @@ type Tag {
...
     typeRegex: String
+    security: [SecurityQuery!]
 }
 
 type Specification {
     description: String
     tags: [Tag!]!
     type: String!
     configuration: ObjectNode!
+    security: [Role!]!
```
nahguam, line 73: Unfortunately the IT is failing and I think it's because of a mismatch between this being a list and the java in `Specification` being a map.

This is because the return types in the graphql layer are just using the model class and not defining their own types in the graphql layer. This makes it impossible to have a clear separation between the graphql and service layers.

The correct answer is to introduce models into the graphql layer - which would mean mostly a copy & paste except for the difference here with an associated converter. This will greatly grow the scope of the change.

A (short-term?) compromise might be to change the current model to a list (making sure to preserve role uniqueness business logic). This does pollutes the model and introduces tech debt but keeps the scope of the PR smaller.

repository/kafka/pom.xml
andykruth-ii: Actually removing all the guava deps and using the hashmap approach to remove downstream issues of an additional dependency.

state/graphql-sender/pom.xml
andykruth-ii: OK - good point. I am changing to this everywhere and removing the guava deps where I added them for consistency.

graphql/api/src/main/resources/stream-registry.graphql
andykruth-ii: OK - renaming done.

graphql/api/src/main/resources/stream-registry.graphql
andykruth-ii: nullables added where appropriate

graphql/api/src/main/resources/stream-registry.graphql
andykruth-ii: OK - now using List<Role> in the model Specification and GraphQL inputs. Keeping Map style in state models.

cli/src/main/java/com/expediagroup/streamplatform/streamregistry/cli/command/Apply.java
```diff
@@ -71,12 +78,23 @@
...
     protected ObjectNode configuration;
+    @Option(names = "--security", converter = ObjectNodeConverter.class)
+    protected ObjectNode security = mapper.createObjectNode();
 
     @Override
     public List<Event<?, ?>> events() {
       return Collections.singletonList(specification(getEntityOptions().key(), getSpecification()));
     }
 
+    public Map<String, List<Principal>> convertSecurityMap(ObjectNode security) {
```
nahguam, line 89: I thought we said there'd be a wrapper around the role, so `Map<Role, List<Principal>>`. @teabot ?

graphql/api/src/main/resources/stream-registry.graphql
```diff
@@ -33,24 +33,52 @@ type Tag {
...
+    role: String
+    principals: [PrincipalInput!]!
+}
+
+input SecurityQuery {
+    roleRegex: String!
+    principalRegex: String!
+}
+
+type Security {
```
nahguam, line 50: In the graphql layer, I think it makes sense to keep the raw strings, otherwise it's adding unnecessary bloat for the API.

i.e.

```
type Security {
  role: String!
  principals: [String!]!
}
```

I realise we will hit the same problem that we did for List vs Map. This will affect the model data structures and it'll have to be `Map<String, List<String>>` in the service model but `Map<Role, List<Principal>>` in lower layers. @teabot comments?

graphql/api/src/main/resources/stream-registry.graphql
```diff
@@ -33,24 +33,52 @@ type Tag {
...
+    role: String
```
nahguam, line 51:
```suggestion
    role: String!
```

graphql/api/src/main/resources/stream-registry.graphql
```diff
@@ -33,24 +33,52 @@ type Tag {
 
 ############ Specification ############
 
+input PrincipalInput {
+    name: String
+}
+
+input SecurityInput {
+    role: String
```
nahguam, line 41:
```suggestion
    role: String!
```

cli/src/main/java/com/expediagroup/streamplatform/streamregistry/cli/command/Apply.java
teabot: That was my understanding also

graphql/api/src/main/resources/stream-registry.graphql
teabot: A `Principal` is a complex type though: `user/U`, `group/G`. Certainly we could serialise to a string (as I've done in those examples), but we might be more likely to edges no?

nahguam: This is true, though currently the `Principal` is just modelled as a `@Value` type. We'd also need to check on the state of graphql input union types or `OneOf` types - quick google later - still open but making progress - it's a rabbit hole of competing PRs https://github.com/graphql/graphql-spec/issues/488

nahguam: maybe at the graphql layer we separate out to something like:

```
type Security {
  role: String!
  users: [String!]!
  groups: [String!]!
}
```

but collapse it down to `Map<Role, List<Principal>>`.

teabot: I was thinking:
```
class Principal {
  PrincipalType type;
  String id;
}

enum PrincipalType {
  USER, GROUP
}
```

cli/src/main/java/com/expediagroup/streamplatform/streamregistry/cli/command/Apply.java
andykruth-ii: @nahguam @teabot - the CLI converts the incoming entities directly into the state model. The state model uses `string -> List<Principal>` because of the restriction in Avro for string keys in Maps.

graphql/api/src/main/resources/stream-registry.graphql
andykruth-ii: @nahguam - I talked with @teabot and we decided to leave the Principal as just a Value instead of a more complex type to avoid schema/enum complexity issues down the road. I have made the change you original suggested to keep the input Principals as a list of strings. I like this change also it should make the graphql simpler.

graphql/api/src/main/resources/stream-registry.graphql
andykruth-ii: handled

graphql/api/src/main/resources/stream-registry.graphql
andykruth-ii: handled

graphql/api/src/main/resources/stream-registry.graphql
```diff
@@ -33,24 +33,48 @@ type Tag {
...
 ############ Specification ############
 
+input SecurityInput {
+    role: String!
+    principals: [String!]!
+}
+
+input SecurityQuery {
+    roleRegex: String!
+    principalRegex: String!
```
nahguam, line 43: On the query side, they should not be mandatory. E.g. you may be searching for an entity with a particular role so not supplying a principalRegex.

```suggestion
    roleRegex: String
    principalRegex: String
```

graphql/api/src/main/resources/stream-registry.graphql
```diff
@@ -33,24 +33,48 @@ type Tag {
 
 ############ Specification ############
 
+input SecurityInput {
+    role: String!
+    principals: [String!]!
```
nahguam, line 38: Wasn't this `PrincipalInput` before? I think `SecurityInput` and `Security` should be complementary.

```
input PrincipalInput {
    name: String
}

input SecurityInput {
    role: String
    principals: [PrincipalInput!]!
}

type Principal {
    name: String
}

type Security {
    role: String
    principals: [Principal!]!
}
```

or 

```
input SecurityInput {
    role: String
    principals: [String!]!
}

type Security {
    role: String
    principals: [String!]!
}
```

andykruth-ii: Done

nahguam (approved):

PR #295, opened 2022-01-10
STREAMP-5897

andykruth-ii:
adding ability to specify arbitrary properties file to event sender/receivers

- [x] CHANGELOG.md updated
- [x] Reviewer assigned


state/kafka-receiver/src/main/java/com/expediagroup/streamplatform/streamregistry/state/kafka/KafkaEventReceiver.java
```diff
@@ -184,5 +209,6 @@ public void close() {
     @NonNull String topic;
     @NonNull String schemaRegistryUrl;
     @NonNull String groupId;
+    String propertiesPath;
```
jamespfaulkner, line 212: Question. Could this be changed to a `Map<String, Object>` and a bean created that reads the properties? Reason I ask is that people would then be able to create a Sender/Receiver with overridden params without having to create a file on disk. Also has the minor benefit would be the reading the file code isn't repeated.

Feel free to push back.

jamespfaulkner (approved):

PR #297, opened 2022-06-17
Stream Registry state metrics

jamespfaulkner:
# stream-registry PR

This PR introduces the ability to add basic metrics to `EntityView`. This is a backwards compatible change.

### Added
* `EntityViews::defaultEntityView` can be used to create `EntityView` that works exactly the same as before.
* `EntityView::meteredEntityView` can be used to create an `EntityView` that exposes minimal metrics.

### Changed
* `EntityViewUpdater` is now an interface with two implementations. This was never exposed outside of it's package so is backwards compatible.

# PR Checklist Forms

- [ ] CHANGELOG.md updated
- [ ] Reviewer assigned

### DIFF
--- a/state/core/src/main/java/com/expediagroup/streamplatform/streamregistry/state/DefaultEntityViewUpdater.java
+++ b/state/core/src/main/java/com/expediagroup/streamplatform/streamregistry/state/DefaultEntityViewUpdater.java
@@ -1,0 +1,116 @@
+/**
+ * Copyright (C) 2018-2022 Expedia, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.expediagroup.streamplatform.streamregistry.state;
+
+import com.expediagroup.streamplatform.streamregistry.state.model.Entity;
+import com.expediagroup.streamplatform.streamregistry.state.model.event.Event;
+import com.expediagroup.streamplatform.streamregistry.state.model.event.SpecificationDeletionEvent;
+import com.expediagroup.streamplatform.streamregistry.state.model.event.SpecificationEvent;
+import com.expediagroup.streamplatform.streamregistry.state.model.event.StatusDeletionEvent;
+import com.expediagroup.streamplatform.streamregistry.state.model.event.StatusEvent;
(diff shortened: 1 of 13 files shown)


state/core/src/main/java/com/expediagroup/streamplatform/streamregistry/state/DefaultEntityViewUpdater.java
```diff
@@ -0,0 +1,116 @@
...
+    } else if (event instanceof SpecificationDeletionEvent) {
+      return delete((SpecificationDeletionEvent<K, S>) event);
+    } else if (event instanceof StatusDeletionEvent) {
+      return delete((StatusDeletionEvent<K, S>) event);
+    } else {
+      throw new IllegalArgumentException("Unknown event " + event);
+    }
+  }
+
+  public <K extends Entity.Key<S>, S extends Specification> Optional<Entity<K, S>> purge(K key) {
```
andykruth-ii, line 58: Should this be an @Override

state/core/src/main/java/com/expediagroup/streamplatform/streamregistry/state/EntityViews.java
```diff
@@ -0,0 +1,75 @@
...
+  @RequiredArgsConstructor(access = PACKAGE)
+  static final class MeteredEntityViewUpdater implements EntityViewUpdater {
+    @NonNull
+    private final EntityViewUpdater delegate;
+    @NonNull
+    private final MeterRegistry meterRegistry;
+
+    @Override
+    public <K extends Entity.Key<S>, S extends Specification> Entity<K, S> update(Event<K, S> event) {
+      meterRegistry.counter("stream_registry_state.receiver.update", Tags.of("event", simpleName(event))).increment();
```
andykruth-ii, line 57: Could we add more tags on these? Possible Key type and the specification type field. It's a shame that URNs are not part of the OOS project, they would ultimately be very useful here. I guess we could implement that in a more advanced metered updater in other projects.

jamespfaulkner: I've extended tags to include the event type and the key. I've not added more than that (key names etc) since it could greatly increase the cardinality, which in turn could increase costs for some observability platforms (e.g. DataDog is costs per unique metric name/tag value)

state/example/src/main/java/com/expediagroup/streamplatform/streamregistry/state/example/ExampleAgentApp.java
```diff
@@ -53,22 +51,22 @@ EventSender eventSender(ApolloClient apolloClient) {
...
+      .groupId(groupId)
+      .schemaRegistryUrl(schemaRegistryUrl)
+      .build();
     return new KafkaEventReceiver(receiverConfig);
   }
 
   @Bean
   EntityView entityView(EventReceiver eventReceiver) {
-    return new DefaultEntityView(eventReceiver);
+    return EntityViews.defaultEntityView(eventReceiver);
```
techabstraction, line 70: Minor: Wonder if we should include a comment to show how to make a metered `EntityView`?

PR #301, opened 2022-07-21
blocking stream deletes with dependent processes

andykruth-ii:


core/src/main/java/com/expediagroup/streamplatform/streamregistry/core/services/StreamService.java
```diff
@@ -106,6 +113,27 @@ public List<Stream> findAll(Predicate<Stream> filter) {
   public void delete(Stream stream) {
     handlerService.handleDelete(stream);
 
+    // Find all Processes that have multiple different stream inputs/outputs which also have this Stream as an input or output.
+    // These processes would be left invalid if the stream was deleted, so block the delete of this Stream.
+    processView.findAll(process ->
+      process.getInputs().stream().anyMatch(input -> input.getStream().equals(stream.getKey())) ||
+        process.getOutputs().stream().anyMatch(output -> output.getStream().equals(stream.getKey()))
```
jamespfaulkner, line 120: I think this might be easier to read if we extract this common log to an method..

```java
  private java.util.stream.Stream<Process> allProcessesForStream(Stream stream) {
    return processView.findAll(process ->
      process.getInputs().stream().anyMatch(input -> input.getStream().equals(stream.getKey())) ||
        process.getOutputs().stream().anyMatch(output -> output.getStream().equals(stream.getKey()))
    );
  }
```

Then the two usages might be slightly more readable...
```java
    // Find all Processes that have multiple different stream inputs/outputs which also have this Stream as an input or output.
    // These processes would be left invalid if the stream was deleted, so block the delete of this Stream.
    allProcessesForStream(stream)
      .filter(process -> processInputOutputStreamKeySet(process).size() > 1)
      .findAny()
      .ifPresent(process -> {
        throw new IllegalStateException("Cannot delete Stream, Processes depend on it including: " + processKeyString(process));
      });

    // Assuming the above check passed, then cascade deletes to Processes which have ONLY this stream as input/output.
    // These types of Processes would be entirely useless without the Stream and so can be safely deleted.
    // This will cascade to ProcessBinding also.
    allProcessesForStream(stream)
      .filter(process -> processInputOutputStreamKeySet(process).size() == 1)
      .forEach(processService::delete);
```

andykruth-ii: yup, done

jamespfaulkner (approved):

PR #302, opened 2022-07-25
adding a release github action

andykruth-ii:


.github/workflows/build.yml
```diff
@@ -1,6 +1,12 @@
 name: Build
 
-on: [push, pull_request]
+on:
+  push:
+    branches-ignore:
+      - master
```
techabstraction, line 6: if we are skipping tests on the release pipeline then we should run them here for master builds?

.github/workflows/release.yml
```diff
@@ -0,0 +1,40 @@
...
+          gpg-passphrase: GPG_PASSPHRASE # env variable for GPG private key passphrase
+          settings-path: ${{ github.workspace }}
+      - name: Cache local Maven repository
+        uses: actions/cache@v2.1.6
+        with:
+          path: ~/.m2
+          key: ${{ runner.os }}-m2-${{ hashFiles('**/pom.xml') }}
+          restore-keys: ${{ runner.os }}-m2
+      - name: Release artifacts
+        run: mvn -Dresume=false release:prepare release:perform -Djib.skip -DskipTests -Darguments="-s $GITHUB_WORKSPACE/settings.xml"
```
techabstraction, line 36: set release version and development version here?

.github/workflows/build.yml
andykruth-ii: reverting back to always doing a build.

.github/workflows/release.yml
andykruth-ii: done

jamespfaulkner (approved):

PR #347, opened 2024-01-19
fix: Adding configurable property to handle double delete events correctly

ojhagaurov:
Configurable property to enable or disable checking if entity exists in EntityView during delete.
We need a way of servicing deletes when the EntityView thinks they are already deleted.

Details

Currently, the OSS project will not allow us to delete entities that don't exist in the EntityView. For example, the implementation of delete from ConsumerMutationImpl includes an ifPresent check:

```
@Override
  public Boolean delete(ConsumerKeyInput key) {
    consumerView.get(key.asConsumerKey()).ifPresent(consumerService::delete);
    return true;
  }
```
  
Eg: When no tombstone for the original schema key id during handling of double delete events, the identified entities will not exist in the EntityView and hence it will be impossible to delete them.

Changed
entityView.exist.check.enabled property to decide if entity exists in EntityView during delete operation.

# PR Checklist Forms

- [ ] CHANGELOG.md updated
- [ ] Reviewer assigned

### DIFF
--- a/graphql/api/src/test/java/com/expediagroup/streamplatform/streamregistry/graphql/mutation/impl/ConsumerBindingMutationImplTest.java
+++ b/graphql/api/src/test/java/com/expediagroup/streamplatform/streamregistry/graphql/mutation/impl/ConsumerBindingMutationImplTest.java
@@ -15,6 +15,7 @@
  */
 package com.expediagroup.streamplatform.streamregistry.graphql.mutation.impl;
 
+import static org.junit.Assert.assertTrue;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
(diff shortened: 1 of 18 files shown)


graphql/api/src/test/java/com/expediagroup/streamplatform/streamregistry/graphql/mutation/impl/ConsumerBindingMutationImplTest.java
```diff
@@ -0,0 +1,89 @@
...
+
+  @Test
+  public void deleteWithCheckExistEnabled() {
+    ReflectionTestUtils.setField(consumerBindingMutation, "checkExistEnabled", true);
+    ConsumerBindingKeyInput key = getConsumerBindingInputKey();
+    when(consumerBindingView.get(any())).thenReturn(Optional.of(getConsumer(key)));
+    Boolean result = consumerBindingMutation.delete(key);
+    verify(consumerBindingService, times(1)).delete(any());
+    verify(consumerBindingView, times(1)).get(any());
+    assert (result);
```
jamespfaulkner, line 62: Minor: Should use JUnit/Hamcrest assertions in tests rather than `assert` keyword (which is typically used in production code, similar to Preconditions)

graphql/api/src/test/java/com/expediagroup/streamplatform/streamregistry/graphql/mutation/impl/ConsumerBindingMutationImplTest.java
```diff
@@ -0,0 +1,89 @@
...
+
+  private ConsumerBindingMutationImpl consumerBindingMutation;
+
+  @Before
+  public void before() throws Exception {
+    consumerBindingMutation = new ConsumerBindingMutationImpl(consumerBindingService, consumerBindingView);
+  }
+
+  @Test
+  public void deleteWithCheckExistEnabled() {
```
jamespfaulkner, line 55: Should we also have the case here where `checkExistEnabled = true`, but there isn't an existing entity, ensuring that delete is not called?

jamespfaulkner (approved):

PR #358, opened 2024-06-11
feat: STREAMP-12186: adding disable flags for entity status support

andykruth-ii:
Adding flags to disable Entity Status.

* Implemented an old TODO to separate repository save functions into two that will allow specification and status saving to be done separately. This will help in disabling entity status support
* Updated all Entity Service Classes to use these distinct save methods
* Updated all Entity Mutation Implementations to make the updateStatus mutation a NOOP which does not persist the status when a toggle is disabled
* Added a convenience constructor to all Entity Model classes that does not require a status value.
* Updated the DefaultEntityView, DefaultEntityViewUpdater and KafkaEventReceiver with option disable toggle for entity status


core/src/main/java/com/expediagroup/streamplatform/streamregistry/core/services/ConsumerBindingService.java
```diff
@@ -1,5 +1,5 @@
 /**
```
andykruth-ii, line 1: All entity service classes are updated to use separated `saveSpecification` and `saveStatus` repository methods.

graphql/api/src/main/java/com/expediagroup/streamplatform/streamregistry/graphql/mutation/impl/ConsumerBindingMutationImpl.java
```diff
@@ -77,7 +80,12 @@ public Boolean delete(ConsumerBindingKeyInput key) {
   @Override
   public ConsumerBinding updateStatus(ConsumerBindingKeyInput key, StatusInput status) {
     ConsumerBinding consumerBinding = consumerBindingView.get(key.asConsumerBindingKey()).get();
-    return consumerBindingService.updateStatus(consumerBinding, status.asStatus()).get();
+
+    if (entityStatusEnabled) {
+      return consumerBindingService.updateStatus(consumerBinding, status.asStatus()).get();
+    } else {
+      return consumerBinding;
+    }
```
andykruth-ii, line 88: All of the individual entity mutation implementation classes have this new Spring value toggle, set to true by default which is the current mode of operation.

When set to false an updateStatus mutation will NOT update the entity status via the entity service, it will just be a successful NOOP.

graphql/api/src/test/java/com/expediagroup/streamplatform/streamregistry/graphql/InputHelper.java
```diff
@@ -0,0 +1,28 @@
...
+
+import com.expediagroup.streamplatform.streamregistry.graphql.model.inputs.StatusInput;
+
+public class InputHelper {
+  private static final ObjectMapper mapper = new ObjectMapper();
+
+  public static StatusInput statusInput() {
+    return StatusInput.builder().agentStatus(mapper.createObjectNode()).build();
+  }
+}
```
andykruth-ii, line 28: Just a convenience test class to generate a statusInput

graphql/api/src/test/java/com/expediagroup/streamplatform/streamregistry/graphql/mutation/impl/ConsumerBindingMutationImplTest.java
```diff
@@ -100,6 +101,39 @@ public void deleteWithCheckExistDisabledWhenEntityDoesNotExist() {
...
+    when(consumerBindingView.get(any())).thenReturn(consumerBinding);
+
+    ConsumerBinding result = consumerBindingMutation.updateStatus(key, statusInput);
+
+    verify(consumerBindingView, times(1)).get(key.asConsumerBindingKey());
+    verify(consumerBindingService, never()).updateStatus(consumerBinding.get(), statusInput.asStatus());
+    assertEquals(consumerBinding.get(), result);
+  }
+
   private ConsumerBindingKeyInput getConsumerBindingInputKey() {
```
andykruth-ii, line 137: Each entity has these two additional tests to test that when entity status is enabled (default) the entity service is called to update the status. The second test is for when entity status is disabled, the entity service should NOT be called when an updateStatus mutation is received.

repository/api/src/main/java/com/expediagroup/streamplatform/streamregistry/repository/Repository.java
```diff
@@ -20,8 +20,16 @@
...
+   */
   T save(T entity);
 
+  T saveSpecification(T entity);
+
+  T saveStatus(T entity);
+
   Optional<T> findById(ID id);
 
   List<T> findAll();
```
andykruth-ii, line 35: Finally handling a TODO from long ago, deprecating the `save` method in favor of two separate `saveSpecification` and `saveStatus` methods.

state/core/src/main/java/com/expediagroup/streamplatform/streamregistry/state/DefaultEntityView.java
```diff
@@ -47,10 +47,19 @@ public class DefaultEntityView implements EntityView {
...
+  }
+
   public DefaultEntityView(EventReceiver receiver) {
     this(receiver, new ConcurrentHashMap<>());
   }
 
+
   @Override
   public CompletableFuture<Void> load(@NonNull EntityViewListener listener) {
     val future = new CompletableFuture<Void>();
```
andykruth-ii, line 65: Adding an optional `entityStatusEnabled` flag to all constructor patterns will can be used to disable entity status persistence.

repository/kafka/src/main/java/com/expediagroup/streamplatform/streamregistry/repository/kafka/DefaultRepository.java
```diff
@@ -48,9 +48,19 @@ abstract class DefaultRepository<
...
   @Override
   public ME save(ME entity) {
+    throw new UnsupportedOperationException();
+  }
+
+  @Override
+  public ME saveSpecification(ME entity) {
     Optional<ME> existing = findById(entity.getKey());
     Entity<SK, SS> stateEntity = converter.convertEntity(entity);
     List<CompletableFuture<Void>> futures = new ArrayList<>();
```
jamespfaulkner, line 66: Do we need these lists of futures now? Could we do similar to what we do in the agents...

```java
  private void sendSync(Event<SK, SS> event) {
    sender.send(event).join();
  }
```

Usages then become:
```java
sendSync(Event.specification(stateEntity.getKey(), stateEntity.getSpecification()));
```

jamespfaulkner (commented):
Couple of questions:

Rather than having multiple instances of the `entityStatusEnabled ` in all instances of the `*MutationImpl` classes, would it be simpler to just ignore these in the `KafkaEventSender`? `enableStatuses` could be added to `Config` there, the sending could then be changed to something like:

```java 
  @Override
  public <K extends Entity.Key<S>, S extends Specification> CompletableFuture<Void> send(@NonNull Event<K, S> event) {
    if (!config.enableStatuses && (event instanceof StatusEvent<K, S> || event instanceof StatusDeletionEvent<K, S>)) {
      return CompletableFuture.completedFuture(null);
    }

    val avroEvent = converter.toAvro(event);
    return send(avroEvent.getKey(), avroEvent.getValue());
  }
```

We'd then remove the if/else in all `*MutationImpl`.

--------

Second question, and maybe this is progress for a follow up. Would it be possible to change the `Event` class hierarchy in allow easier differentiation between entity/status events. 

I'm thinking two new interfaces with `SpecificationEvent`/`StatusEvent`/`SpecificationDeletionEvent`/`StatusDeletionEvent` extending the correct interface.
```java 
public interface EntityEvent<K extends Entity.Key<S>, S extends Specification> extends Event<K, S> { }

public interface StatusEvent<K extends Entity.Key<S>, S extends Specification> extends Event<K, S> { }
```

We could then have a new `EventReceiverListener`  which deals only with the entity events (meaning agents really are shielded from Status events...

```java
<K extends Entity.Key<S>, S extends Specification> void onEvent(EntityEvent<K, S> event);
```

repository/kafka/src/main/java/com/expediagroup/streamplatform/streamregistry/repository/kafka/DefaultRepository.java
```diff
@@ -48,9 +48,19 @@ abstract class DefaultRepository<
   private final Converter<ME, MK, SK, SS> converter;
   private final Class<SK> stateKeyClass;
 
-  //TODO deprecate - split specification from status so they are saveable separately.
+  /**
+   * @deprecated Use {link {@link #saveSpecification(com.expediagroup.streamplatform.streamregistry.model.Entity)}} or
+   * {link {@link #saveStatus(com.expediagroup.streamplatform.streamregistry.model.Entity)}}
+   * to save the entity specification or entity status independently.
+   */
+  @Deprecated
```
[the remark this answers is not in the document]
andykruth-ii: yep good call, done

repository/kafka/src/main/java/com/expediagroup/streamplatform/streamregistry/repository/kafka/DefaultRepository.java
andykruth-ii: yep, missed that. done

andykruth-ii:
@jamespfaulkner - OK summary of updates:
* Added the entity status toggle to KafkaEventSender similar to KafkaEventReceiver
* Better tests for KafkaEventReceiver
* Removed all MutationImpl changes, no longer necessary
* Removed all EntityView changes, no longer necessary
* removed repository.send method instead of deprecating so that we get a compile error instead of runtime error
* removed futures on single event send in saveSpecification

state/kafka-receiver/src/main/java/com/expediagroup/streamplatform/streamregistry/state/kafka/KafkaEventReceiver.java
```diff
@@ -38,12 +38,8 @@
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.atomic.*;
 
-import lombok.Builder;
-import lombok.NonNull;
-import lombok.RequiredArgsConstructor;
-import lombok.Value;
+import lombok.*;
```
jamespfaulkner, line 41: Minor: star import

jamespfaulkner (approved):

jamespfaulkner (approved):

PR #361, opened 2024-06-20
fix: STREAMP-12336: removing validation on schema key if it's changing in update request

sunlnu:
# stream-registry PR

_&lt;So what happens currently, if customer want to change the SchemaKey for a stream, we are allowing to do this.
Although we are not updating schemaKey and in control-plane stream still pointing to older schema.
Probably, we should not allow to customer to change the schemaKey for a stream, and give the valid error if any customer trying to do. 
For now plan is, not throw error from OOS project, in place of we will throw error from egsp-stream-registry.
&gt;_

### Added
* _&lt;detail item of what was added&gt;_
* _&lt;describe functionality added&gt;_

### Changed
* _&lt;schema validation for a stream has been changed&gt;_

### Deleted
* _&lt;detail item of what was removed&gt;_

# PR Checklist Forms

- [ ] CHANGELOG.md updated
- [ ] Reviewer assigned

### DIFF
--- a/core/src/main/java/com/expediagroup/streamplatform/streamregistry/core/services/StreamService.java
+++ b/core/src/main/java/com/expediagroup/streamplatform/streamregistry/core/services/StreamService.java
@@ -82,8 +82,6 @@ public class StreamService {
     if (existing.isEmpty()) {
       throw new ValidationException("Can't update " + stream.getKey() + " because it doesn't exist");
     }
-    System.out.println(existing.get().getSchemaKey());
-    System.out.println(stream.getSchemaKey());
     if (stream.getSchemaKey() == null) {
       stream.setSchemaKey(existing.get().getSchemaKey());
     }
--- a/core/src/main/java/com/expediagroup/streamplatform/streamregistry/core/services/StreamService.java
+++ b/core/src/main/java/com/expediagroup/streamplatform/streamregistry/core/services/StreamService.java
@@ -82,9 +82,14 @@ public class StreamService {
     if (existing.isEmpty()) {
       throw new ValidationException("Can't update " + stream.getKey() + " because it doesn't exist");
     }
+
     if (stream.getSchemaKey() == null) {
       stream.setSchemaKey(existing.get().getSchemaKey());
+    } else if (!existing.get().getSchemaKey().equals(stream.getSchemaKey())) {
+      throw new ValidationException("Stream = " + stream.getKey() + " update failed, because existing schemaKey = " +
+        existing.get().getSchemaKey() + " is not matching with given schemaKey = " + stream.getSchemaKey());
     }
+
     streamValidator.validateForUpdate(stream, existing.get());
     stream.setSpecification(handlerService.handleUpdate(stream, existing.get()));
     return saveSpecification(stream);
--- a/core/src/main/java/com/expediagroup/streamplatform/streamregistry/core/services/StreamService.java
+++ b/core/src/main/java/com/expediagroup/streamplatform/streamregistry/core/services/StreamService.java
@@ -82,6 +82,11 @@ public class StreamService {
     if (existing.isEmpty()) {
       throw new ValidationException("Can't update " + stream.getKey() + " because it doesn't exist");
     }
+
+    if (stream.getKey() != null && !existing.get().getKey().equals(stream.getKey())) {
+      throw new ValidationException("can't update " + stream.getKey() + "because existing schema key = " +
+        existing.get().getKey() + " is not matching with given schema = " + stream.getKey());
+    }
     stream.setSchemaKey(existing.get().getSchemaKey());
     streamValidator.validateForUpdate(stream, existing.get());
     stream.setSpecification(handlerService.handleUpdate(stream, existing.get()));
--- a/core/src/main/java/com/expediagroup/streamplatform/streamregistry/core/services/StreamService.java
+++ b/core/src/main/java/com/expediagroup/streamplatform/streamregistry/core/services/StreamService.java
@@ -83,7 +83,7 @@ public class StreamService {
       throw new ValidationException("Can't update " + stream.getKey() + " because it doesn't exist");
     }
 
-    if (stream.getKey() != null && !existing.get().getKey().equals(stream.getKey())) {
+    if (stream.getSchemaKey() != null && !existing.get().getSchemaKey().equals(stream.getSchemaKey())) {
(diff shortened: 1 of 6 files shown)


core/src/main/java/com/expediagroup/streamplatform/streamregistry/core/validators/StreamValidator.java
```diff
@@ -57,11 +58,15 @@ private void requireExistingDomain(Stream stream) {
...
-    if (stream.getSchemaKey() == null) {
-      throw new ValidationException("Schema must be specified");
-    }
+    requireSchemaSpecified(stream);
     if (!schemaView.exists(stream.getSchemaKey())) {
       throw new ValidationException("Schema does not exist");
     }
   }
+
+  private void requireSchemaSpecified(Stream stream) {
```
sunlnu, line 67: Doing this validation from egsp-stream-reg for update case.
`!schemaView.exists(stream.getSchemaKey())`

it/src/test/java/com/expediagroup/streamplatform/streamregistry/it/StreamTestStage.java
```diff
@@ -79,18 +79,6 @@ public void upsert() {
...
 
-    try {
-      SchemaKeyInput nonExisting = SchemaKeyInput.builder()
-          .domain(factory.domainName)
-          .name("nonExisting")
-          .build();
-      client.getOptionalData(factory.upsertStreamMutationBuilder()
-          .schema(nonExisting)
-          .build()).get();
-    } catch (RuntimeException ex) {
```
sunlnu, line 90: Doing this validation from egsp-stream-reg for update case.

core/src/main/java/com/expediagroup/streamplatform/streamregistry/core/services/StreamService.java
```diff
@@ -82,7 +82,9 @@ public Optional<Stream> update(Stream stream) throws ValidationException {
     if (existing.isEmpty()) {
       throw new ValidationException("Can't update " + stream.getKey() + " because it doesn't exist");
     }
-    stream.setSchemaKey(existing.get().getSchemaKey());
+    if (stream.getSchemaKey() == null) {
```
sunlnu, line 85: this check for SMS, if update request not having schema .

core/src/main/java/com/expediagroup/streamplatform/streamregistry/core/validators/StreamValidator.java
```diff
@@ -57,11 +54,15 @@ private void requireExistingDomain(Stream stream) {
...
+    if (stream.getSchemaKey() == null) {
+      throw new ValidationException("Schema must be specified");
+    }
+  }
```
andykruth-ii, line 67: What is the goal of the changes in these validations? As far as I can tell all that has been changed is that previously, on update a schema key was required to be present and the schema exists. But now, it seems like on update the schema is required to be present, but is NOT required to exist? I don't think this is desirable?

core/src/test/java/com/expediagroup/streamplatform/streamregistry/core/services/StreamServiceTest.java
```diff
@@ -182,6 +183,35 @@ public void update() {
...
+  @Test(expected = ValidationException.class)
+  public void updateWithChangedSchemaKey() {
+    StreamKey key = new StreamKey();
+    key.setDomain("domain");
+    key.setName("stream");
+    key.setVersion(1);
+    Stream existingEntity = new Stream(key, new Specification(), new SchemaKey("domain", "stream_v1"));
+    Stream updatedEntity = new Stream(key, new Specification(), new SchemaKey("domain", "stream_v2"));
+    when(streamRepository.findById(key)).thenReturn(Optional.of(existingEntity));
+    streamService.update(updatedEntity);
```
andykruth-ii, line 195: Should this throw an error?

core/src/test/java/com/expediagroup/streamplatform/streamregistry/core/services/StreamServiceTest.java
```diff
@@ -182,6 +183,35 @@ public void update() {
...
+  @Test
+  public void updateWithSchemaKeyNull() {
+    StreamKey key = new StreamKey();
+    key.setDomain("domain");
+    key.setName("stream");
+    key.setVersion(1);
+    Stream existingEntity = new Stream(key, new Specification(), new SchemaKey("domain", "stream_v1"));
+    Stream updatedEntity = new Stream(key, new Specification(), null);
+    when(streamRepository.findById(key)).thenReturn(Optional.of(existingEntity));
+    streamService.update(updatedEntity);
```
andykruth-ii, line 208: can we validate the output contains the non null schemakey

it/src/test/java/com/expediagroup/streamplatform/streamregistry/it/StreamTestStage.java
```diff
@@ -1,5 +1,5 @@
 /**
- * Copyright (C) 2018-2021 Expedia, Inc.
```
andykruth-ii, line 2: Are changes to this file still necessary?

core/src/test/java/com/expediagroup/streamplatform/streamregistry/core/services/StreamServiceTest.java
sunlnu: yeah i have added this 
`@Test(expected = ValidationException.class)`
do you want me to add error message ??

it/src/test/java/com/expediagroup/streamplatform/streamregistry/it/StreamTestStage.java
sunlnu: yeah because now, we have check if schema is not matching with existing schemaKey.

core/src/test/java/com/expediagroup/streamplatform/streamregistry/core/services/StreamServiceTest.java
sunlnu: added

core/src/test/java/com/expediagroup/streamplatform/streamregistry/core/services/StreamServiceTest.java
sunlnu: now asserting the exact error .

andykruth-ii (approved):

